PR #13, opened 2024-03-13
feat: match ignores against directly invoked paths

seren5240:
for https://github.com/getgrit/rewriter/issues/9042

Unfortunately, there isn't a way to do this out of the box with `ignore`'s interface. There was a crate exposing [closer functionality](https://github.com/nathankleyn/gitignore.rs?tab=readme-ov-file) but it's been archived.

### DIFF
--- a/crates/language/src/target_language.rs
+++ b/crates/language/src/target_language.rs
@@ -389,8 +389,7 @@ fn find_ignore_files(start_path: &Path, name: &str) -> Result<Vec<PathBuf>> {
         current_path = parent.to_path_buf();
     }
 
-    let root_path = current_path;
-    for entry in WalkDir::new(root_path).into_iter().filter_map(|e| e.ok()) {
+    for entry in WalkDir::new(start_path).into_iter().filter_map(|e| e.ok()) {
         let path = entry.path();
         if path.file_name() == Some(std::ffi::OsStr::new(name)) {
             gitignore_files.push(path.to_path_buf());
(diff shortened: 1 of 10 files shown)


crates/language/src/target_language.rs
```diff
@@ -369,11 +373,60 @@ impl PatternLanguage {
...
 }
 
+#[cfg(feature = "finder")]
+fn find_ignore_files(start_path: &Path, name: &str) -> Result<Vec<PathBuf>> {
+    use path_absolutize::Absolutize;
+
+    let mut gitignore_files = Vec::new();
+
+    let mut current_path = start_path.absolutize()?.canonicalize()?;
+    while let Some(parent) = current_path.parent() {
```
morgante, line 383: We should probably stop once we hit a git repo boundary.

crates/language/src/target_language.rs
```diff
@@ -369,11 +373,60 @@ impl PatternLanguage {
...
+
+    Ok(gitignore_files)
+}
+
+#[cfg(feature = "finder")]
+fn is_file_ignored(path: &Path) -> Result<bool> {
+    use ignore::{gitignore::GitignoreBuilder, Match};
+    use path_absolutize::Absolutize;
+
+    let git_ignores = find_ignore_files(path, ".gitignore")?;
```
morgante, line 406: We'll want to memoize the .gitignore finding. It's very expensive to do it for every file.

morgante (changes requested):

morgante:
Thanks for adding the memo-ization, but I'm still a bit hesitant about the additional complexity in this PR. We are re-implementing key parts of ignore (directory traversal upwards / searching for gitignore files) to work around what feels like a bug.

Doing a bit more research:
* https://github.com/BurntSushi/ripgrep/issues/829 looks related / symptomatic of the problem

PR #15, opened 2024-03-13
fix: wasm feature doesn't break build

ilevyor:
hides some wasm specific features behind arch wasm32, so they don't break the build due to conflicting feature flags


morgante:
What build is broken? Tests all pass on main.

morgante:
We don't use workspace builds for anything though.

Ideally we should do this via feature flags not architecture selectors.

ilevyor:
> We don't use workspace builds for anything though.
> 
> Ideally we should do this via feature flags not architecture selectors.

mutually exclusive features are an anti-pattern: https://doc.rust-lang.org/cargo/reference/features.html
Do you not build locally?

morgante:
> Do you not build locally?

I do. `yarn build` works fine.

morgante (approved):

PR #38, opened 2024-03-18
chore: warn instead of forbid unused crate

ilevyor:
tests are compiled independently so can result in false positives when enforcing this lint, warn is sufficient to fail CI.


morgante (approved):
This is fine, but let's stop editing this. It's not productive.

PR #61, opened 2024-03-22
fix: graceful error handling for `grit check` in uninitialized contexts

mparker3:
(hopefully!) no functional changes. 

calling `grit check` before `grit init` right now panics:
```
❯ marzano check
thread '<unnamed>' panicked at /Users/mattparker/code/gritql/crates/gritcache/src/new_cache.rs:43:65:
called `Result::unwrap()` on an `Err` value: Failed to open cache file

Caused by:
    No such file or directory (os error 2)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

Checking no_println_in_lsp
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 0/1thread 'main' panicked at /Users/mattparker/code/gritql/crates/cli/src/commands/check.rs:197:58:
called `Result::unwrap()` on an `Err` value: sending on a closed channel
```
    
fix results in:
```
    ❯ marzano check
Error: Failed to read cache file /Users/mattparker/.cargo/.grit/.gritmodules/mismatches_cache

Caused by:
    No such file or directory (os error 2)
```
    
 probably not a super common occurence but it _was_ one of the first things I ran into test driving the CLI, can replicate with the added test in the absence of the fix. Happy to take this PR in a different direction if needed

## Summary by CodeRabbit

- **Refactor**
	- Simplified error handling in cache operations to enhance reliability.
	- Improved testing for cache behavior when encountering missing files.

### DIFF
--- a/crates/gritcache/src/new_cache.rs
+++ b/crates/gritcache/src/new_cache.rs
@@ -66,14 +66,10 @@ impl ThreadedCache {
         let file_vector = match std::fs::read(path) {
             Ok(bytes) => bytes,
             Err(e) => {
-                if e.kind() == std::io::ErrorKind::NotFound {
-                    return Ok(HashMap::new());
-                } else {
-                    return Err(e).context(format!(
-                        "Failed to read cache file {}",
-                        path.to_string_lossy()
-                    ));
-                }
+                return Err(e).context(format!(
+                    "Failed to read cache file {}",
+                    path.to_string_lossy()
+                ));
             }
         };
 
@@ -182,6 +178,9 @@ mod tests {
         if mismatches_cache_path.exists() {
             std::fs::remove_file(&mismatches_cache_path)?;
         }
+
+        // Assert cache raises err on no file existing + no cache refresh 
+        assert!(ThreadedCache::new(path.clone(), false).await.is_err());
 
         // Create an empty cache
         let (cache, manager) = ThreadedCache::new(path.clone(), true).await?;
--- a/crates/gritcache/src/new_cache.rs
+++ b/crates/gritcache/src/new_cache.rs
@@ -39,8 +39,8 @@ impl ThreadedCache {
         };
 
         let (sender, receiver) = mpsc::channel::<HashKey>();
+        let mut writer = Self::new_writer(&mismatches_path)?;
         let manager = thread::spawn(move || {
-            let mut writer = Self::new_writer(&mismatches_path).unwrap();
             while let Ok(key) = receiver.recv() {
                 writer.write_all(&key).unwrap();
             }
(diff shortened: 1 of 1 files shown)


morgante (changes requested):
Thanks for contributing! This doesn't look quite right, since if the cache doesn't exist we actually want to initialize an empty cache so it could be saved afterwards.

Can you adjust this so that's preserved (and add a test confirming it)?

mparker3:
playing with this a little more -- the creation logic for the backing file is correct as implemented assuming exists on the FS, so this definitely isn't the right fix.

would a better approach be to add a check [here]((https://github.com/mparker3/gritql/blob/c8f9160197e68aa5340a45f19005706a13a60f30/crates/gritcache/src/paths.rs#L35) that the `cache_dir` in question exists + bail if not w/ a message to `grit init`? (vs. trying to do `.grit`/`.gritmodules` initialization ad-hoc)

morgante:
Yeah, I think we want to do it this way:
- If the directories don't exist, fail gracefully with a message about the cache
- If the directories exist (`grit init` ran), initialize like we did here.

> would a better approach be to add a check [here](https://github.com/mparker3/gritql/blob/c8f9160197e68aa5340a45f19005706a13a60f30/crates/gritcache/src/paths.rs#L35) that the cache_dir in question exists + bail if not w/ a message to grit init? (vs. trying to do .grit/.gritmodules initialization ad-hoc)

Generally we should avoid doing pre-emptive checks but instead handle failures if they come up. I think it might actually be fine to [edit this](https://github.com/getgrit/gritql/blob/4173cc996d6920d35759163e5ba607b6656bab98/crates/gritcache/src/new_cache.rs#L43) so instead of panicking that thread gracefully terminates with a suggestion to run `grit init` first.

crates/gritcache/src/new_cache.rs
```diff
@@ -39,8 +39,8 @@ impl ThreadedCache {
         };
 
         let (sender, receiver) = mpsc::channel::<HashKey>();
+        let mut writer = Self::new_writer(&mismatches_path)?;
```
mparker3, line 42: @morgante let me know if there's a reason not to do this i'm missing - seems more straightforward to bail during creation than to do so from the writer thread

morgante: This looks fine - if the compiler is happy, I'm happy.

morgante (approved):

PR #66, opened 2024-03-24
feat: `Context` trait

arendjr:
This PR renames the `Context` struct to `MarzanoContext` and introduces a new `Context` trait to take its place. While this PR doesn't accomplish much on its own, the end goal is to be able to extract the patterns and matching engine from `marzano-core` and put them into a crate of their own (`grit-core-patterns`?). This crate could in turn be used by Biome for its plugin engine.

There is one change in this PR that might be slightly controversial: I've refactored `builtin_functions.rs` a bit (see the [first commit](https://github.com/getgrit/gritql/commit/746fc0adedc95216b6dbda08f0a5cfe0472ff386)) so that context is no longer passed to built-in functions. This creates no issues now, but could possibly create limitations with future functions. But the reason I did this is that otherwise the `Builtins` type would need to have a generic argument for the context, and because the context can return the built-ins, this created a recursive dependency. This got me into a fight with the borrow-checker I was not able to resolve otherwise.

If this PR is approved, I will continue decoupling stuff until I'm able to extract the `grit-core-patterns` crate without having dependencies on Tree-Sitter or other Marzano crates. Please let me know if there are concerns with this strategy.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `Context` trait with comprehensive methods for enhanced pattern, predicate, and function handling.
	- Added new implementations and adjustments for `MarzanoContext` to align with the `Context` trait.
- **Refactor**
	- Reorganized the module structure for better clarity and usability.
	- Promoted certain modules to public to increase their accessibility.
	- Shifted towards using trait objects and trait bounds for more flexible context handling across various pattern-related functionalities.
	- Improved parameter handling in `suppress.rs` and `text_unparser.rs` for more efficient string processing.


ilevyor:
> There is one change in this PR that might be slightly controversial: I've refactored builtin_functions.rs a bit (see the [first commit](https://github.com/getgrit/gritql/commit/746fc0adedc95216b6dbda08f0a5cfe0472ff386)) so that context is no longer passed to built-in functions.

Unfortunately this is somewhat of an issue. The context struct is only ever used (or at the very least introduced) for built in functions. The functions they are used in are in a closed source ai_builtins crate.

crates/core/src/pattern/built_in_functions.rs
```diff
@@ -91,12 +89,10 @@ pub struct BuiltInFunction {
 impl BuiltInFunction {
     fn call<'a>(
         &self,
-        args: &'a [Option<Pattern>],
```
morgante, line 94: We have some cloud functions that require the Pattern (not the resolved pattern) so this change won't work for them.

arendjr:
> Unfortunately this is somewhat of an issue. The context struct is only ever used (or at the very least introduced) for built in functions. The functions they are used in are in a closed source ai_builtins crate.

I’ll try if I can come up with a different solution to this then!

arendjr:
I've updated the PR so that `Builtins` doesn't need a generic argument and remains Marzano-specific. It means Biome will have to make its own implementation, but I think that should be fine.

morgante (approved):

morgante:
@arendjr This is going to need more work:
- Clippy isn't happy
- Rust builds fail

arendjr:
WASM builds should be fixed now.

The Clippies weren't related to my PR, but it looks like they simply only showed up with WASM builds. Fixed too.

morgante:
@arendjr FYI you don't have to listen to the AI rabbit unless you want to. We're just trying it out - so far it seems to mostly add noise.

PR #69, opened 2024-03-25
feat: add language option.

### LINKED ISSUE #58: `--language` flag on the CLI
For simple inline patterns, needing to specify the language in the declaration is a hindrance.

For example:
```
$ grit apply 'language css
r"--red-(.+)"($var)'
```

This would be more ergonomic:
```
$ grit apply 'r"--red-(.+)"($var)'' --language css
```

Shylock-Hg:
Close #58 
/claim #58 

## Summary by CodeRabbit

- **New Features**
	- Added support for specifying a language option when applying patterns through the CLI. This allows users to override the default language specified in the pattern.

- **Tests**
	- Introduced new tests to ensure the correct application of patterns with the language option across various scenarios.

- **Refactor**
	- Enhanced pattern language functionality and CLI argument parsing with new dependencies and implementations.


morgante (changes requested):
Thanks for contributing! Please provide a few integration tests, both for inline patterns and named patterns.

Also, if the pattern has a defined language that differs from the specified language it should probably be an error. `--language` just changes the default.

crates/cli/src/commands/apply_pattern.rs
```diff
@@ -109,6 +109,8 @@ pub struct ApplyPatternArgs {
     /// Clear cache before running apply
     #[clap(long = "refresh-cache", conflicts_with = "cache")]
     pub refresh_cache: bool,
+    #[clap(long = "language")]
```
morgante, line 112:
```suggestion
    #[clap(long = "language", short="lang")]
```

morgante: You can make it an alias instead.

morgante: Clap supports command aliases. Make 'lang' an alias for language.

crates/cli_bin/tests/apply.rs
```diff
@@ -2015,3 +2015,101 @@ fn ignores_file_in_grit_dir() -> Result<()> {
...
+    // Keep _temp_dir around so that the tempdir is not deleted
+    let (_temp_dir, dir) = get_fixture("simple_python", false)?;
+
+    // from the tempdir as cwd, run init
+    run_init(&dir.as_path())?;
+
+    // from the tempdir as cwd, run marzano apply
+    let mut apply_cmd = get_test_cmd()?;
+    apply_cmd.current_dir(dir.as_path());
+    // apply_cmd.current_dir(basic_path);
```
ilevyor, line 2030: no commented out code

crates/cli_bin/tests/apply.rs
```diff
@@ -2015,3 +2015,101 @@ fn ignores_file_in_grit_dir() -> Result<()> {
...
+    apply_cmd.arg("apply").arg("--force").arg("pattern.grit").arg("--language").arg("python");
```
ilevyor, line 2031: we only want this feature for inline patterns. If the pattern is defined in a file we should log a warning that the option was unused

crates/cli_bin/tests/apply.rs
```diff
@@ -2015,3 +2015,101 @@ fn ignores_file_in_grit_dir() -> Result<()> {
...
+
+    // assert that it matches snapshot
+    assert_snapshot!(content);
+
+    Ok(())
+}
+
+#[test]
+fn language_option_conflict_apply() -> Result<()> {
+    let pattern = r"language php
```
ilevyor, line 2086: lets use a currently supported language. I know were adding php soon, but it's not in yet

crates/cli/src/commands/apply_pattern.rs
```diff
@@ -109,6 +109,8 @@ pub struct ApplyPatternArgs {
     /// Clear cache before running apply
     #[clap(long = "refresh-cache", conflicts_with = "cache")]
     pub refresh_cache: bool,
+    #[clap(long = "language", alias="lang")]
+    pub language: Option<String>,
```
morgante, line 113: Use an enum, so documentation generation and clap will work better.

morgante: Only certain values are valid here, so you should use an enum.

morgante: You can use this existing enum: https://github.com/getgrit/gritql/blob/51863c285b717f58a9138390986246082d5d41c4/crates/language/src/target_language.rs#L46

crates/language/src/target_language.rs
```diff
@@ -371,6 +374,61 @@ impl PatternLanguage {
     }
 }
 
+impl ValueEnum for PatternLanguage {
```
morgante, line 377: I think you could just derive this. See this example: https://github.com/clap-rs/clap/issues/4327

crates/language/src/target_language.rs
```diff
@@ -371,6 +374,61 @@ impl PatternLanguage {
...
+            PatternLanguage::Sql,
+            PatternLanguage::Vue,
+            PatternLanguage::Toml,
+            PatternLanguage::Universal,
+        ]
+    }
+
+    fn to_possible_value(&self) -> Option<PossibleValue> {
+        match self {
+            PatternLanguage::JavaScript => Some(PossibleValue::new("js(js_do_not_use)")),
```
morgante, line 406: We should `#skip` this so it's not a valid option.

crates/language/src/target_language.rs
```diff
@@ -371,6 +374,61 @@ impl PatternLanguage {
...
+            PatternLanguage::TypeScript => Some(PossibleValue::new("js(typescript)")),
+            PatternLanguage::Tsx => Some(PossibleValue::new("js")),
+            PatternLanguage::Html => Some(PossibleValue::new("html")),
+            PatternLanguage::Css => Some(PossibleValue::new("css")),
+            PatternLanguage::Json => Some(PossibleValue::new("json")),
+            PatternLanguage::Java => Some(PossibleValue::new("java")),
+            PatternLanguage::CSharp => Some(PossibleValue::new("csharp")),
+            PatternLanguage::Python => Some(PossibleValue::new("python")),
+            PatternLanguage::MarkdownBlock => Some(PossibleValue::new("markdown(block)")),
```
morgante, line 415: Just `markdown`, and skip the inline variant.

crates/language/src/target_language.rs
```diff
@@ -371,6 +374,61 @@ impl PatternLanguage {
...
+            PatternLanguage::Go => Some(PossibleValue::new("go")),
+            PatternLanguage::Rust => Some(PossibleValue::new("rust")),
+            PatternLanguage::Ruby => Some(PossibleValue::new("ruby")),
+            PatternLanguage::Solidity => Some(PossibleValue::new("solidity")),
+            PatternLanguage::Hcl => Some(PossibleValue::new("hcl")),
+            PatternLanguage::Yaml => Some(PossibleValue::new("yaml")),
+            PatternLanguage::Sql => Some(PossibleValue::new("sql")),
+            PatternLanguage::Vue => Some(PossibleValue::new("vue")),
+            PatternLanguage::Toml => Some(PossibleValue::new("toml")),
+            PatternLanguage::Universal => Some(PossibleValue::new("universal")),
```
morgante, line 426: This isn't a valid target language, so it should be excluded.

crates/cli/src/commands/apply_pattern.rs
```diff
@@ -285,6 +293,22 @@ pub(crate) async fn run_apply_pattern(
                 }
             }
         };
+        if let Some(lang_option) = &arg.language {
+            if is_pattern_file(&pattern) {
```
morgante, line 297: I don't think we need to special case pattern files - we should also not allow overriding the `language` in those cases if there's a conflict.

Similarly I should be able to `grit apply` a `foo.grit` that doesn't have a `language` specified.

Concretely, the purpose of the `--language` option is to change the *default* pattern language from JS to whatever is passed.

crates/cli_bin/tests/apply.rs
```diff
@@ -2015,3 +2015,167 @@ fn ignores_file_in_grit_dir() -> Result<()> {
...
+        "Command didn't finish successfully: {}",
+        String::from_utf8(output.stderr)?
+    );
+
+    // Read back the main.py file
+    let target_file = dir.join("main.py");
+    let content: String = std::fs::read_to_string(target_file)?;
+
+    // assert that it matches snapshot
+    assert_snapshot!(content);
```
morgante, line 2046: No snapshot needed here, just assert that it's unchanged.

crates/cli_bin/tests/apply.rs
```diff
@@ -2015,3 +2015,167 @@ fn ignores_file_in_grit_dir() -> Result<()> {
...
+        !output.status.success(),
+        "Command with conflict language option and pattern should fail"
+    );
+
+    // Read back the main.py file
+    let target_file = dir.join("main.py");
+    let content: String = std::fs::read_to_string(target_file)?;
+
+    // assert that it matches snapshot
+    assert_snapshot!(content);
```
morgante, line 2146: You can skip the snapshot here.

```suggestion
```

morgante (changes requested):

crates/language/src/target_language.rs
```diff
@@ -371,6 +374,61 @@ impl PatternLanguage {
...
+            PatternLanguage::Ruby,
+            PatternLanguage::Solidity,
+            PatternLanguage::Hcl,
+            PatternLanguage::Yaml,
+            PatternLanguage::Sql,
+            PatternLanguage::Vue,
+            PatternLanguage::Toml,
+            PatternLanguage::Universal,
+        ]
+    }
```
ilevyor, line 402: this function already exists see `enumerate` on line 308. feel free to make it return a static ref instead of a vector.
```suggestion
```

crates/cli_bin/tests/apply.rs
morgante: Understood. Instead of a snapshot you should just confirm the current contents match the original contents.

crates/cli_bin/tests/apply.rs
morgante: A snapshot requires maintenance, and doesn't express the intent as clearly. Asserting the final contents against the original comments better captures the intent.

morgante: You're not listening - you should *NOT* use a snapshot here. A snapshot is unnecessary if all you're asserting is that the content hasn't changed!

crates/cli_bin/tests/apply.rs
```diff
@@ -2015,3 +2015,170 @@ fn ignores_file_in_grit_dir() -> Result<()> {
...
+        "Command didn't finish successfully: {}",
+        String::from_utf8(output.stdout)?
+    );
+
+    // Read back the main.py file
+    let target_file = dir.join("main.py");
+    let content: String = std::fs::read_to_string(target_file)?;
+
+    // assert that it matches snapshot
+    assert_snapshot!(content);
```
morgante, line 2078: As I said before, you do not need a snapshot here. Please listen.

morgante (changes requested):
Do not use snapshots incorrectly.

crates/cli_bin/tests/apply.rs
morgante: Tests look good now actually.

crates/cli/src/commands/apply_pattern.rs
morgante: This is the only remaining change. I'm the tech lead, you don't need @ilevyor to approve.

morgante (approved):
Thanks for contributing!

Issue #71, opened 2024-03-25
Empty field matching is unintuitive


morgante:
In JSON, empty strings in snippets become wildcard matches:

```
engine marzano(0.1)
language json

`"x": ""`
```

This will match `"x": ""` but also `"x": "foo"`. When we find an empty node in a snippet we probably want to restrict it to only match other empty nodes.

PR #74, opened 2024-03-26
fix: json empty string only matches empty strings

ilevyor:
fixes: https://github.com/getgrit/gritql/issues/71
by adding a function to the language trait which lists fields which must be matched regardless of weather or not they are present in the snippet. ordinarily fields that are not present are ignored.

## Summary by CodeRabbit

- **New Features**
	- Enhanced pattern matching capabilities to include handling of mandatory empty arguments.
	- Introduced new test scenarios for improved JSON parsing and matching accuracy.
	- Added support for identifying mandatory empty fields in JSON structures.

- **Refactor**
	- Modified internal logic for processing pattern arguments and JSON field sorting to enhance functionality.

- **Tests**
	- Implemented new tests to ensure accurate matching behavior for JSON strings.

### DIFF
--- a/crates/core/src/pattern/patterns.rs
+++ b/crates/core/src/pattern/patterns.rs
@@ -505,7 +505,7 @@ impl Pattern {
                 )?));
             }
             let fields: &Vec<Field> = &node_types[sort as usize];
-            let args = fields
+            let mut args = fields
                 .iter()
                 .filter(|field| {
                     node.child_by_field_id(field.id()).is_some()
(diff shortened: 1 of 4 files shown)


crates/core/src/pattern/patterns.rs
```diff
@@ -562,6 +562,26 @@ impl Pattern {
...
+                    (
+                        field.id(),
+                        true,
+                        Pattern::List(Box::new(List::new(Vec::new()))))
+                } else {
+                    (
+                        field.id(),
+                        false,
+                        Pattern::Dynamic(DynamicPattern::Snippet(DynamicSnippet {
+                        parts: vec![DynamicSnippetPart::String("".to_string())],
```
morgante, line 578: Isn't this special casing it as a `""`? Not all fields would have an empty string as a valid option, let alone using this syntax.

Why not use `Constant::null` instead?

ilevyor: This mirrors what we use for the `.` pattern. I think having a proper dot pattern would probably be better, but I don't think it's worth changing here.

morgante: Ok, we did have https://github.com/getgrit/gritql/blob/0cb34abd869ee9121d7689c77f3a0ede2ea4c21e/crates/core/src/pattern/patterns.rs#L98 but this is fine?

PR #83, opened 2024-03-27
chore: use paths in `FileOwners` and `Binding::FileName`

arendjr:
This uses `PathBuf`/`&Path` instead of `String`/`&str` in the `FileOwners` struct and the `Binding::FileName` variant. I had to do them both at once because of a pointer equality check in `apply_effects()` which would fail if an intermediate conversion had been done in between.

This introduces a few conversions here and there, but this seemed preferred compared to spreading the use of paths further across the codebase. Using paths has a few advantages:

* It will improve compatibility with systems that don't use UTF8 for file paths, notably Windows.
* It makes it clearer to developers what is a path vs. a regular string. There are a few places in the codebase where it currently isn't clear whether a file path or a file's source is expected.
* Rust provides helpers for dealing with paths, which are only available on the `Path` types.
* Other Rust software, including Biome, typically expect `Path`/`PathBuf` to be used when dealing with files, so this will ease interoperability.

## Summary by CodeRabbit

- **Refactor**
	- Enhanced file handling by transitioning to path-based representations for improved robustness.
	- Improved comparison logic for file names to ensure accuracy in processing.
	- Updated methods and structures to support path-based file manipulation, enhancing system flexibility and efficiency.

### DIFF
--- a/crates/core/src/binding.rs
+++ b/crates/core/src/binding.rs
@@ -507,9 +507,7 @@ impl<'a> Binding<'a> {
             Binding::Node(source, _) => Some(source),
             Binding::String(source, _) => Some(source),
             Binding::List(source, _, _) => Some(source),
-            // maybe should be none?
-            Binding::FileName(source) => source.to_str(),
-            Binding::ConstantRef(_) => None,
+            Binding::FileName(..) | Binding::ConstantRef(..) => None,
         }
     }
 
--- a/crates/core/src/binding.rs
+++ b/crates/core/src/binding.rs
@@ -7,6 +7,7 @@ use marzano_language::target_language::TargetLanguage;
 use marzano_util::analysis_logs::{AnalysisLogBuilder, AnalysisLogs};
 use marzano_util::position::{Position, Range};
 use std::ops::Range as StdRange;
+use std::path::Path;
 use std::{borrow::Cow, collections::HashMap, fmt::Display};
 use tree_sitter::Node;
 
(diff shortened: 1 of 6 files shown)


crates/core/src/pattern/step.rs
```diff
@@ -232,10 +231,9 @@ impl Matcher for Step {
             let file = state.files.get_file(file_ptr);
             let mut match_log = file.matches.borrow_mut();
 
-            let filename_string = &file.name;
+            let filename_path = &file.name;
 
-            // todo shouldn't have to wrap new_filename in const
-            let mut new_filename: Constant = Constant::String(filename_string.to_owned());
+            let mut new_filename = filename_path.clone();
```
arendjr, line 236: I was able to resolve this TODO while I was at it.

crates/core/src/binding.rs
```diff
@@ -506,9 +507,18 @@ impl<'a> Binding<'a> {
             Binding::Node(source, _) => Some(source),
             Binding::String(source, _) => Some(source),
             Binding::List(source, _, _) => Some(source),
-            // maybe should be none?
-            Binding::FileName(source) => Some(source),
-            Binding::ConstantRef(_) => None,
+            Binding::FileName(..) | Binding::ConstantRef(..) => None,
```
arendjr, line 510: Returning `None` (as the comment suggested) would've previously caused potential issues in `step.rs` where the source was compared against a filename. With the introduction of `as_filename`, this should no longer be an issue. At least the tests still pass :)

ilevyor (approved):

PR #85, opened 2024-03-27
chore: introduce `AstNode` trait

arendjr:
This introduces an `AstNode` trait, which gets implemented by a `NodeWithSource` struct. The goal is again to reduce dependencies on TreeSitter, so while `NodeWithSource` still uses a TreeSitter node internally, the `AstNode` trait doesn't expose any TreeSitter internals.

Currently there are only 2 methods implemented, which aid with the implementations of the `Before` and `After` patterns. As can be seen, this also reduces the `match binding` expressions. Later, this should hopefully make it possible to place `Binding` itself behind an abstract trait as well.

## Summary by CodeRabbit

- **New Features**
	- Introduced an advanced representation for AST (Abstract Syntax Tree) nodes, enhancing the way AST nodes and their source code are managed and navigated.
- **Refactor**
	- Improved the clarity and efficiency of node and binding handling in the core logic, particularly in pattern matching and traversal functionalities.
- **Chores**
	- Reorganized the module structure for better clarity and maintenance.

### DIFF
--- a/crates/core/src/ast_node.rs
+++ b/crates/core/src/ast_node.rs
@@ -1,0 +1,47 @@
+use tree_sitter::Node;
+
+/// Represents an AST node and offers convenient AST-specific functionality.
+///
+/// This trait should be free from dependencies on TreeSitter.
+pub trait AstNode: Sized {
+    /// Returns the next node, ignoring trivia such as whitespace.
+    fn next_non_trivia_node(&self) -> Option<Self>;
+
+    /// Returns the previous node, ignoring trivia such as whitespace.
+    fn previous_non_trivia_node(&self) -> Option<Self>;
+}
+
+/// A TreeSitter node, including a reference to the source code from which it
+/// was parsed.
+pub(crate) struct NodeWithSource<'a> {
+    pub node: Node<'a>,
+    pub source: &'a str,
+}
+
+impl<'a> NodeWithSource<'a> {
+    pub fn new(node: Node<'a>, source: &'a str) -> Self {
+        Self { node, source }
+    }
+}
+
+impl<'a> AstNode for NodeWithSource<'a> {
+    fn next_non_trivia_node(&self) -> Option<Self> {
+        let mut current_node = self.node.clone();
+        loop {
+            if let Some(sibling) = current_node.next_named_sibling() {
+                return Some(Self::new(sibling, self.source));
+            }
+            current_node = current_node.parent()?;
+        }
+    }
+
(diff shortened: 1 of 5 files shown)


crates/core/src/binding.rs
```diff
@@ -324,6 +325,39 @@ pub(crate) fn linearize_binding<'a>(
...
+                None
+            }
+        }
+    }
+
+    /// Returns the node from which this binding was created.
+    ///
+    /// This differs from [`Self::as_node()`] in that it returns the node for an
+    /// entire list when the binding applies to a list item.
+    pub(crate) fn get_node(&self) -> Option<NodeWithSource<'a>> {
```
arendjr, line 352: I made these two separate implementations to mirror the original implementations. But I notice if I change all calls to `get_node()` with `as_node()`, the tests also remain green... and the other way around as well. It makes me wonder whether this is simply a case of lacking test coverage, or was this divergence in the implementations an oversight in the original implementation?

morgante: Likely an oversight, we can consolidate on get_node.

morgante (approved):

crates/core/src/ast_node.rs
```diff
@@ -0,0 +1,47 @@
...
+/// Represents an AST node and offers convenient AST-specific functionality.
+///
+/// This trait should be free from dependencies on TreeSitter.
+pub trait AstNode: Sized {
+    /// Returns the next node, ignoring trivia such as whitespace.
+    fn next_non_trivia_node(&self) -> Option<Self>;
+
+    /// Returns the previous node, ignoring trivia such as whitespace.
+    fn previous_non_trivia_node(&self) -> Option<Self>;
+}
```
ilevyor, line 12: Not blocking for this pr,
Ideally this trait should also implement the trait from tree-sitter-traverse. alternatively we could excise tree-sitter-traverse and have this trait replace that as the crate is only one file. tree-sitter-traverse trait is not dependent on tree-sitter

crates/core/src/binding.rs
ilevyor: I would consolidate using get_node, or maybe even return None for lists/empty.  rather than as_node. as_node looks buggy to me for two reasons. `child_by_field_id` gives the first node in the field, so as_node does not distinguish between fields with multiple nodes or single nodes. additionally child_by_field_id (if it behaves like children_by_field_id) includes unnamed nodes which we do not care about while matching

ilevyor (approved):

PR #93, opened 2024-03-27
chore: create `grit-util` crate

arendjr:
This introduces a new crate with the traversal algorithm from `tree-sitter-traversal`. The source of the crate has been integrated so that even the optional TreeSitter dependency could be removed, and the `AstCursor` trait could be tied to our new `AstNode` trait, which has also been put in the same crate.

See previous discussion: https://github.com/getgrit/gritql/pull/85#discussion_r1541353476

## Summary by CodeRabbit

- **New Features**
	- Introduced `AstNode` and `AstCursor` traits for abstract syntax tree (AST) node representation and traversal.
	- Added new modules for handling AST nodes and their traversal, enhancing the library's functionality.
- **Refactor**
	- Updated various modules to use `grit_util` for node traversal, replacing previous implementations.
	- Reorganized imports and adjusted module paths and dependencies across multiple files.
	- Modified function calls and parameters to align with new `grit_util` traversal logic.
- **Chores**
	- Removed the `ast_node` module from public exports, streamlining the library's interface.
	- Added `node_with_source` and `runtime` modules to the public API, expanding the available utilities.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1306,6 +1306,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "grit-util"
+version = "0.1.0"
+
+[[package]]
 name = "grit_cache"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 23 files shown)


ilevyor (approved):
reminds me that cargo check, didn't make it to the new repo CI, should add it back

Issue #84, opened 2024-03-27
Doesn't handle removal of trailing whitespace for python language


jonathan-s:
In the following example https://app.grit.io/studio?key=jdcyl3I8UCOr1F_WxhQ-z we have some valid python code that runs perfectly fine. But because the editor has been set to remove trailing whitespace grit complains that the code is not properly indented, when in fact this is valid python and runs fine as is.

morgante:
We might just need to adjust this: https://microsoft.github.io/monaco-editor/typedoc/interfaces/editor.IGlobalEditorOptions.html#trimAutoWhitespace

It's a problem in the studio (not GritQL) so the CLI should work fine.

ilevyor:
@jonathan-s  Is the problem that the online editor at app.grit.io strips trailing whitespace, or that your personal editor strips the whitespace, I believe
https://github.com/getgrit/gritql/pull/99
should fix

jonathan-s:
My personal editor strips whitespace

PR #99, opened 2024-03-27
fix: handle empty lines for languages with semantic whitespace

ilevyor:
fixes: https://github.com/getgrit/gritql/issues/84
lines containing only whitespace will be rewritten as empty lines

## Summary by CodeRabbit

- **Refactor**
	- Improved modularity and readability of range adjustments in code.
- **Tests**
	- Added tests for Python support with empty lines in code snippets.

### DIFF
--- a/crates/core/src/binding.rs
+++ b/crates/core/src/binding.rs
@@ -103,6 +103,17 @@ fn pad_snippet(padding: &str, snippet: &str) -> String {
     result
 }
 
+fn adjust_ranges(substitutions: &mut [(EffectRange, String)], index: usize, delta: isize) {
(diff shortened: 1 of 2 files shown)


morgante:
I'm not sure we actually want to do this. The root cause of that issue looked like a problem in Monaco, not anything in GritQL.

ilevyor:
> I'm not sure we actually want to do this. The root cause of that issue looked like a problem in Monaco, not anything in GritQL.

the code is valid regardless of the whitespace on the empty line, I ran the python. I'm sure people have empty lines in their codebases in the middle of a code block I don't see why we shouldn't support that.

morgante (approved):

PR #100, opened 2024-03-27
chore: implement `Binding::list_items()`

arendjr:
Implemented `Binding::list_items()` and used it in `Every`, `List` and `Some`.

Also ran into one more case that could be covered by `Binding::singleton()`.

## Summary by CodeRabbit

- **New Features**
	- Introduced the ability to iterate over items in a list through the `list_items` method for enhanced list handling.
- **Refactor**
	- Improved pattern matching logic for better clarity and efficiency, particularly in handling list items.
	- Optimized the handling of bindings and list items across various matchers to ensure accurate pattern execution.

### DIFF
--- a/crates/core/src/binding.rs
+++ b/crates/core/src/binding.rs
@@ -9,6 +9,7 @@ use marzano_util::analysis_logs::{AnalysisLogBuilder, AnalysisLogs};
 use marzano_util::node_with_source::NodeWithSource;
 use marzano_util::position::{Position, Range};
 use marzano_util::tree_sitter_util::children_by_field_id_count;
+use std::iter;
 use std::ops::Range as StdRange;
 use std::path::Path;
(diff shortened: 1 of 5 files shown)


crates/core/src/binding.rs
```diff
@@ -522,6 +523,44 @@ impl<'a> Binding<'a> {
...
+                                }
+                            }
+                            let result = cursor.node();
+                            if !cursor.goto_next_sibling() {
+                                done = true;
+                            }
+                            return Some(result);
+                        }
+                        None
+                    })
```
arendjr, line 551: This function is the implementation of TreeSitter's `children_by_field_id()`. I inlined it because it allows me to `move` the cursor into the closure, or it wouldn't live long enough to be returned.

ilevyor (approved):

PR #104, opened 2024-03-28
chore: move `binding_is_suppressed()` onto `Binding`

arendjr:
Moves `binding_is_suppressed()` to `Binding::is_suppressed()`.

## Summary by CodeRabbit

- **Refactor**
	- Improved suppression check logic by integrating `is_suppressed` directly into the `Binding` object, enhancing code readability and maintainability.


crates/core/src/text_unparser.rs
```diff
@@ -28,9 +27,9 @@ pub(crate) fn apply_effects<'a>(
     current_name: Option<&str>,
     logs: &mut AnalysisLogs,
 ) -> Result<(String, Option<Vec<Range<usize>>>)> {
-    let mut our_effects = Vec::new();
+    let mut our_effects = Vec::with_capacity(effects.len());
```
arendjr, line 30: Tiny optimization :)

morgante (approved):
Also, fine, but now has a merge conflict.

crates/core/src/suppress.rs
```diff
@@ -8,39 +8,41 @@ use crate::binding::Binding;
...
+        lang: &impl Language,
+        current_name: Option<&str>,
+    ) -> Result<bool> {
+        let (src, node) = match self {
+            Self::Node(src, node) => (src, node),
+            Self::String(_, _) => return Ok(false),
+            Self::List(src, node, _) => (src, node),
+            Self::Empty(src, node, _) => (src, node),
+            Self::FileName(_) => return Ok(false),
+            Self::ConstantRef(_) => return Ok(false),
```
ilevyor, line 23: I'm somewhat dissatisfied with binding confining String, and ConstantRef as they don't actually point to a source file, maybe they should be moved under resolved pattern. Just something I've been thinking about. FileName is a bit borderline since it does point to an underlying file, but you can't do think like suppress filename

crates/core/src/text_unparser.rs
ilevyor: unpausing is quite slow in files with lots of rewrites. Most files don't have many rewrites, but those that do can be quite slow. We repeatedly clone many strings, I'm sure there are many low hanging fruits there

ilevyor (approved):

arendjr:
No problem, I’ll resolve the conflict after the weekend!

PR #106, opened 2024-03-28
chore: move `node_text()` to `AstNode::text()`

arendjr:
Turns `node_text()` into a trait method for `AstNode`.

Also found another case that could be simplified with `Binding::singleton()`.

## Summary by CodeRabbit

- **Refactor**
	- Simplified handling and text extraction logic for bindings in the core functionality.
- **New Features**
	- Enhanced pattern matching capabilities for string constants.
- **Enhancements**
	- Introduced new methods for improved text representation of nodes.

### DIFF
--- a/crates/core/src/binding.rs
+++ b/crates/core/src/binding.rs
@@ -3,6 +3,7 @@ use crate::pattern::resolved_pattern::CodeRange;
 use crate::pattern::state::{get_top_level_effects, FileRegistry};
 use crate::pattern::{Effect, EffectKind};
 use anyhow::{anyhow, Result};
+use grit_util::AstNode;
 use marzano_language::language::{FieldId, Language};
 use marzano_language::target_language::TargetLanguage;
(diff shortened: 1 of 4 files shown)


crates/core/src/binding.rs
```diff
@@ -461,7 +462,9 @@ impl<'a> Binding<'a> {
     pub fn text(&self) -> String {
         match self {
             Binding::Empty(_, _, _) => "".to_string(),
-            Binding::Node(source, node) => node_text(source, node).to_string(),
+            Binding::Node(source, node) => {
+                NodeWithSource::new(node.clone(), source).text().to_string()
+            }
```
arendjr, line 467: Not a high priority now, but I think `Binding::Node` can be adjusted to have a `NodeWithSource` instead of separate fields. That'll probably enable a lot of code simplifications down the line.

morgante (approved):

ilevyor (approved):

PR #110, opened 2024-03-28
fix: expose runtime

morgante:
Internal extensions (AI functions) still need this.


morgante:
Not needed after all.

PR #132, opened 2024-04-02
chore: reduce `Binding` variant usages

arendjr:
This PR further reduces the amount of usages of `Binding::*` variants, mostly by using helper functions.

Usages of `Binding::*` variants internal to the struct have been renamed to `Self::*`, so I can quickly filter them out to see which ones remain.

It looks like after this, the only remaining cases are construction of `Binding::List` and `Binding::Empty`. These cannot be abstracted, but that should not be a problem once we have an abstract `Binding` trait, since instances will still be created of the concrete type (probably `MarzanoBinding` then).

## Summary by CodeRabbit

- **Refactor**
	- Enhanced pattern matching and binding creation logic for improved efficiency and readability.
	- Optimized pattern resolution processes to enhance performance.
	- Streamlined text unparser for more accurate output generation.

### DIFF
--- a/crates/core/src/pattern/or.rs
+++ b/crates/core/src/pattern/or.rs
@@ -74,10 +74,10 @@ impl Matcher for Or {
         if let ResolvedPattern::Binding(binding_vector) = &binding {
             for p in self.patterns.iter() {
                 // filter out pattern which cannot match because of a mismatched node type
-                if let (Binding::Node(_src, binding_node), Pattern::ASTNode(node_pattern)) =
-                    (binding_vector.last().unwrap(), p)
+                if let (Some(binding_node), Pattern::ASTNode(node_pattern)) =
+                    (binding_vector.last().and_then(Binding::as_node), p)
                 {
-                    if node_pattern.sort != binding_node.kind_id() {
+                    if node_pattern.sort != binding_node.node.kind_id() {
                         continue;
                     }
                 }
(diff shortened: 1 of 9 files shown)


crates/core/src/pattern/or.rs
```diff
@@ -74,10 +74,10 @@ impl Matcher for Or {
         if let ResolvedPattern::Binding(binding_vector) = &binding {
             for p in self.patterns.iter() {
                 // filter out pattern which cannot match because of a mismatched node type
-                if let (Binding::Node(_src, binding_node), Pattern::ASTNode(node_pattern)) =
-                    (binding_vector.last().unwrap(), p)
+                if let (Some(binding_node), Pattern::ASTNode(node_pattern)) =
+                    (binding_vector.last().and_then(Binding::as_node), p)
```
ilevyor, line 78: Not documented but the binding vector in a resolved pattern should be wrapped in a non-empty-vec struct as the vector is always created with an initial element, and only ever appended to, so last shouldn't really be an optional here .

ilevyor (approved):

ilevyor:
Can you just fix the clippy complaints?

arendjr:
Oh, that's odd. The fix for the Clippy seems pushed, and locally it's indeed gone. Maybe the CI incorrectly cached the result?

PR #133, opened 2024-04-02
refactor: simplify list bindings

arendjr:
This PR simplifies list bindings by getting of the `Lists` enum and the `ListBinding` struct, both in `resolved_pattern.rs`.

This involves a compromise: Instead of lazily retrieving the list items, this eagerly retrieves them when creating the `JoinFn`. The tests seem fine with this, but let me know if this has other repercussions.

## Summary by CodeRabbit

- **Refactor**
	- Updated the `Binding` struct to enhance the return type of `list_items` function.
	- Simplified pattern matching and refactored import statements in built-in functions for better efficiency.
	- Enhanced the handling of `Contains` functionality to streamline the code.
	- Improved the logic for accessing and resolving list items in the `ListIndex` implementation.
	- Revamped the handling of lists in the `JoinFn` struct for improved performance and simplicity.

### DIFF
--- a/crates/core/src/pattern/resolved_pattern.rs
+++ b/crates/core/src/pattern/resolved_pattern.rs
@@ -99,47 +99,21 @@ impl<'a> File<'a> {
 }
 
 #[derive(Debug, Clone, PartialEq)]
-pub(crate) struct ListBinding<'a> {
-    pub(crate) src: &'a str,
-    pub(crate) parent_node: Node<'a>,
-    pub(crate) field: FieldId,
-}
-
-impl<'a> ListBinding<'a> {
-    pub(crate) fn new(src: &'a str, parent_node: Node<'a>, field: FieldId) -> Self {
-        Self {
-            src,
-            parent_node,
-            field,
-        }
-    }
-}
-
-#[derive(Debug, Clone, PartialEq)]
-pub(crate) enum Lists<'a> {
-    Binding(ListBinding<'a>),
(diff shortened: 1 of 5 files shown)


crates/core/src/pattern/resolved_pattern.rs
```diff
@@ -98,48 +98,22 @@ impl<'a> File<'a> {
...
-#[derive(Debug, Clone, PartialEq)]
-pub(crate) enum Lists<'a> {
-    Binding(ListBinding<'a>),
-    Resolved(Vector<ResolvedPattern<'a>>),
-}
-
 #[derive(Debug, Clone, PartialEq)]
 pub struct JoinFn<'a> {
-    pub(crate) list: Lists<'a>,
+    pub(crate) list: Vector<ResolvedPattern<'a>>,
```
ilevyor, line 103: Is this what you referring to where you had to eagerly evaluate? I think this is fine. The purpose of the function structs is to be able to lazily evaluate the text the nodes are pointing to. Specifically if a function is on the right hand side of a rewrite then the function will be evaluated at linearization time. Sometimes the functions are eagerly evaluated eg. the text function eagerly evaluates the enclosed resolved pattern.

ilevyor (approved):

crates/core/src/pattern/resolved_pattern.rs
arendjr: Thanks! This is the one indeed.

PR #141, opened 2024-04-02
fix: use matches instead of equals

ilevyor:
fixes the failing tests on main, and changes CI to use on pull_request instead of pull_request_target

## Summary by CodeRabbit

- **Chores**
	- Updated GitHub Actions workflows to trigger on `pull_request` events for enhanced code quality checks.


arendjr:
I suppose you may need to force-push, since the test will keep failing until the fixed workflow is in `main`.

morgante (approved):

PR #148, opened 2024-04-03
refactor: move node traversal to `AstNode` trait

arendjr:
This PR removes `parent_traverse.rs` in favor of a new `AstNode::ancestors()` method. A similar `AstNode::children()` method was added too.

I've also renamed `next/previous_non_trivia_node()` to `next/previous_named_sibling()`. The non-trivia terminology comes from Biome, and it felt out of place here.

## Summary by CodeRabbit

- **Refactor**
	- Enhanced node traversal logic across various components, improving the way nodes are retrieved and processed.
	- Simplified ancestor traversal in pattern matching by utilizing a more streamlined approach.
	- Updated suppression logic to more accurately handle comments and node suppression conditions.
	- Streamlined AST node handling with new methods for better ancestor and children iteration.
	- Removed outdated modules and reorganized code for better clarity and maintenance.
- **Documentation**
	- Adjusted documentation to reflect changes in node traversal and suppression logic.


crates/core/src/suppress.rs
```diff
@@ -1,34 +1,25 @@
...
-                .chain(ParentTraverse::new(TreeSitterParentCursor::new(
-                    node.clone(),
-                )))
-        {
-            let mut cursor = n.walk();
-            let children = n.children(&mut cursor);
-            for c in children {
-                if !(lang.is_comment(c.kind_id()) || lang.is_comment_wrapper(&c)) {
+        let target_range = node.node.range();
+        for n in node.children().chain(node.ancestors()) {
```
arendjr, line 17: I've left the semantics intact, but to be honest, I still have my doubts about why we would want to chain children and ancestors together. The inner loop also loops over children, so why would we want to check grandchildren of a node for suppression comments?

morgante (approved):

crates/core/src/suppress.rs
```diff
@@ -39,21 +30,18 @@ impl<'a> Binding<'a> {
...
     target_range: &Range,
     current_name: Option<&str>,
     lang: &impl Language,
 ) -> bool {
-    let child_range = comment_node.range();
-    let Ok(text) = comment_node.utf8_text(src.as_bytes()) else {
-        return false;
-    };
+    let child_range = comment_node.node.range();
+    let text = comment_node.text();
```
arendjr, line 39: `NodeWithSource::text()` has the same intention as `Node::utf8_text(src.as_bytes())`, but it doesn't return a `Result`. Theoretically it could panic, but that seems very unlikely because it means the node range is incorrect.

crates/core/src/suppress.rs
morgante: @seren5240 originally implemented this, do you recall why?

Since comment attachment points are somewhat hard to predict in tree-sitter, it does make sense for us to iterate up/down the tree a bit.

crates/grit-util/src/ast_node_traversal.rs
```diff
@@ -41,29 +41,6 @@ pub trait AstCursor {
...
-
-    fn goto_next_sibling(&mut self) -> bool {
-        T::goto_next_sibling(self)
-    }
-
-    fn node(&self) -> Self::Node {
-        T::node(self)
-    }
-}
-
```
arendjr, line 66: This blanket implementation was unused.

crates/core/src/suppress.rs
seren5240: I don't remember, but you can find out by removing it and seeing which unit tests fail.

PR #153, opened 2024-04-04
refactor: use `NodeWithSource` in `Binding` variants

arendjr:
At this point, I bit the bullet and changed the variants of `Binding` that stored separate `node` and `src` values to hold a single `NodeWithSource` instead. The PR is slightly bigger because I had to add a few more traversal methods to cover cases that weren't covered yet (I should have done those in reverse order).

The nice thing is that you can see this PR completely strips the `use tree_sitter::Node;` imports from several of the files it touches. Progress! :)

## Summary by CodeRabbit

- **Refactor**
    - Enhanced node and source handling for improved functionality and clarity.
    - Optimized equivalence checking logic and simplified pattern matching for increased efficiency.
- **New Features**
    - Expanded AST node navigation capabilities with new methods for efficient structure traversal.
- **Style**
    - Implemented `Debug` trait and refined comparisons for better debugging support.

### DIFF
--- a/crates/core/src/binding.rs
+++ b/crates/core/src/binding.rs
@@ -9,12 +9,9 @@ use marzano_language::target_language::TargetLanguage;
 use marzano_util::analysis_logs::{AnalysisLogBuilder, AnalysisLogs};
 use marzano_util::node_with_source::NodeWithSource;
 use marzano_util::position::{Position, Range};
-use marzano_util::tree_sitter_util::children_by_field_id_count;
-use std::iter;
 use std::ops::Range as StdRange;
 use std::path::Path;
 use std::{borrow::Cow, collections::HashMap, fmt::Display};
-use tree_sitter::Node;
 
 // the inner references hold the mutable state
 #[derive(Debug, Clone)]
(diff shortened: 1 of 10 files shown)


crates/core/src/equivalence.rs
```diff
@@ -9,41 +10,38 @@ impl<'a> Binding<'a> {
...
-                            .all(|(node1, node2)| are_equivalent(source1, &node1, source2, &node2))
-                }
+            Self::List(parent_node1, field1) => match other {
+                Self::List(parent_node2, field2) => parent_node1
+                    .children_by_field_id(*field1)
+                    .zip_longest(parent_node2.children_by_field_id(*field2))
+                    .all(|zipped| match zipped {
+                        EitherOrBoth::Both(node1, node2) => are_equivalent(&node1, &node2),
+                        EitherOrBoth::Left(_) | EitherOrBoth::Right(_) => false,
+                    }),
```
arendjr, line 34: Before, we compared `children_by_field_id_count()` of both nodes in a separate step, which meant we would often end up iterating the children twice: Once for counting, and once for comparing. Using `zip_longest()`, we can do both in one iteration.

arendjr:
Looks like I introduced a bug downstream. I’ll try to create a test case for this issue in the core create too. Where can I find the failing test case?

crates/core/src/binding.rs
```diff
@@ -546,34 +528,11 @@ impl<'a> Binding<'a> {
...
-                        None
-                    })
-                    .filter(|child| child.is_named())
-                    .map(|named_child| NodeWithSource::new(named_child, src)),
-                )
-            }
+            Self::List(parent_node, field_id) => Some(
+                parent_node
+                    .children_by_field_id(*field_id)
+                    .filter(|item| item.node.is_named()),
```
ilevyor, line 534: ideally `children_by_field_id` should just filter already. probably worth checking this doesn't break anything though

crates/core/src/binding.rs
```diff
@@ -587,18 +546,21 @@ impl<'a> Binding<'a> {
...
 
     pub fn is_truthy(&self) -> bool {
         match self {
             Self::Empty(..) => false,
-            Self::List(_, node, field_id) => {
-                let child_count = children_by_field_id_count(node, *field_id);
+            Self::List(node, field_id) => {
+                let child_count = node
+                    .children_by_field_id(*field_id)
+                    .filter(|child| child.node.is_named())
```
ilevyor, line 562: this filter comes up everywhere, should be built into the utility

crates/util/src/node_with_source.rs
```diff
@@ -14,6 +18,27 @@ impl<'a> NodeWithSource<'a> {
     pub fn new(node: Node<'a>, source: &'a str) -> Self {
         Self { node, source }
     }
+
+    pub fn children_by_field_id(&self, field_id: u16) -> impl Iterator<Item = Self> + Clone {
+        ChildrenByFieldIterator::new(self, field_id)
+    }
```
ilevyor, line 24: ideally we add a `named_children_by_field_id` as it comes up so often, could then also get rid of the tree-sitter-util file.

ilevyor (commented):
Looks good, One of the patterns in the standard library is failing you can clone here https://github.com/getgrit/stdlib and run `<path_to_bin> patterns test --exclude ai` to recreate locally

crates/util/src/node_with_source.rs
arendjr: Haha, yes, I was already thinking cleaning up `tree_sitter_util` will be my next PR :)

crates/core/src/binding.rs
arendjr: I now have `children_by_field_id()` and `named_children_by_field_id()`. Using `named_children_by_field_id()` everywhere indeed resulted in some regressions.

crates/core/src/binding.rs
```diff
@@ -316,7 +311,7 @@ impl<'a> Binding<'a> {
     }
 
     pub(crate) fn from_node(node: NodeWithSource<'a>) -> Self {
-        Self::Node(node.source, node.node)
+        Self::Node(node)
     }
```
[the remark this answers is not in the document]
arendjr: I'll keep it for a separate PR though, so I don't bloat this one any more.

arendjr:
Should be fixed now. Turns out I shot myself in the foot with poor naming in the previous PR and got functions returning next/previous *siblings* mixed up with next/previous *nodes* (may be siblings, but can be other relatives).

I tried creating another test case for it, but to honest I found it hard to see which part of the stdlib pattern triggered the error case, so I'm going to leave it at this. At least CI did already catch the error.

ilevyor (approved):

PR #164, opened 2024-04-05
refactor: separate pattern factory from patterns

arendjr:
Here's one that might be a little more controversial: I anticipate that all the `PatternX::from_node()` methods will always remain tied to TreeSitter, because they construct patterns from concrete `Node` types. I could maybe simplify them a little by using `NodeWithSource`, but I'm afraid it will be infeasible to make them abstract enough to use `impl AstNode` later on.

So instead, I propose to split the factory functions from the patterns and create a `pattern_factory` module for them, which is what this PR does. I've only created factory functions for `Accessor` and `Step` so far, and I've also moved the compiler (which directly uses TreeSitter of course) into the `pattern_factory` directory. I've moved `auto_wrap.rs` too, since it's only used by `compiler.rs` and `step.rs`.

This creates a bit of a circular dependency between the `pattern` module and the `pattern_factory` module, but that should be temporary. If we agree on this approach, I'll continue moving the functions out of the other patterns until `pattern` doesn't depend on `pattern_factory` anymore. And of course, the end goal is still for `pattern` to not depend on TreeSitter at all anymore, so eventually it can be moved wholesale into the future `grit-core-patterns` crate.

PS.: This PR appears large, but it's really almost all updated import statements and a little bit of moving around of code.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new module `pattern_factory` to enhance pattern handling capabilities.
  - Added constants for index values and default file names to improve standardization across patterns.

- **Refactor**
  - Streamlined import statements across various pattern-related files for improved clarity and maintainability.
  - Restructured the `Accessor` struct and renamed `Key` to `AccessorKey` for better code readability.
  - Consolidated imports under `super` and `pattern_factory` to maintain consistency in the codebase.
  - Simplified and refactored the implementation of the `Step` struct, removing unnecessary pattern checks.

- **Style**
  - Reorganized imports and adjusted module structures across numerous files to align with coding standards and enhance code organization.


morgante:
@arendjr Please request review when this is ready (or mark as draft).

arendjr:
I don’t think GitHub lets me request reviews by handle on this repo (though I can request re-review when someone’s left one already), but it’s ready for review, @morgante.

morgante (approved):

PR #185, opened 2024-04-08
WIP: Convert all node compilers

arendjr:
This PR shows how I intend to convert the remaining node compilers. This one has all the node compilers converted, using a few techniques:

* Added a `NodeCompilationContext` to bundle the various arguments passed to node compilers.
* Added a `from_node_with_rhs()` trait method that can be used to pass `is_rhs` when desired.
* As a bonus, all node compilers now take a `NodeWithSource` instead of `Node`. This, plus, the `NodeCompilationContext` enabled a *lot* of code simplifications.

Of course this PR is too large now and the tests still fail, so I’ll start cutting out some smaller pieces and submit those for review instead.

cc @morgante


morgante:
This approach looks reasonable, happy to review a PR switching the signatures over.

ilevyor (approved):

PR #195, opened 2024-04-10
chore: migrate pattern compilers

arendjr:
Takes another bite out of https://github.com/getgrit/gritql/pull/185 and converts the `Pattern::from_node()` to `PatternCompiler`. Also includes the `MapCompiler` and `ListCompiler`, as well as tiny compilers for all the constant types and ranges.

After this, only snippet handling and calls should be left, though they represent a significant chunk as well.

## Summary by CodeRabbit

- **New Features**
	- Enhanced pattern compilation with the introduction of specialized compiler modules for boolean, float, integer, string constants, lists, maps, and ranges.
	- Improved handling and creation of AST nodes for pattern compilation, supporting more complex pattern types with named arguments.
- **Refactor**
	- Significant refactoring across pattern and compiler modules to streamline pattern creation and compilation processes.
	- Simplified several structs by removing methods that parsed nodes, focusing instead on direct value handling.
- **Documentation**
	- Expanded the public interface to include new compiler modules, making it easier for developers to extend and utilize pattern compilation features.


crates/core/src/pattern/ast_node.rs
```diff
@@ -47,27 +53,30 @@ impl ASTNode {
...
-            if named_args.len() > 1 {
-                return Err(anyhow!("comment node has more than one field"));
-            }
-            if named_args.is_empty() {
-                return Ok(Self::new(sort, args));
-            }
-            let (name, node) = &named_args[0];
-            if *name != "content" {
-                return Err(anyhow!("unknown field name {} for comment node", name));
+            match named_args.len().cmp(&1) {
```
arendjr, line 56: I changed this a little bit to avoid another clone, at which point Clippy suggested I turn it into a `match`. It does look cleaner this way :)

ilevyor (approved):

PR #203, opened 2024-04-11
chore: move `CodeRange` to `grit-util` crate

arendjr:
This makes it easier to create `CodeRange`s from abstract `AstNode`s.

## Summary by CodeRabbit

- **Refactor**
	- Enhanced code efficiency and clarity in core pattern matching and text parsing functionalities.
- **New Features**
	- Improved precision in text processing by integrating advanced code range handling.

### DIFF
--- a/crates/core/src/binding.rs
+++ b/crates/core/src/binding.rs
@@ -1,9 +1,8 @@
 use crate::inline_snippets::inline_sorted_snippets_with_offset;
-use crate::pattern::resolved_pattern::CodeRange;
 use crate::pattern::state::{get_top_level_effects, FileRegistry};
 use crate::problem::{Effect, EffectKind};
 use anyhow::{anyhow, Result};
-use grit_util::AstNode;
+use grit_util::{AstNode, CodeRange};
 use marzano_language::language::{FieldId, Language};
 use marzano_language::target_language::TargetLanguage;
 use marzano_util::analysis_logs::{AnalysisLogBuilder, AnalysisLogs};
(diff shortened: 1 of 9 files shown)


crates/core/src/binding.rs
```diff
@@ -392,6 +385,7 @@ impl<'a> Binding<'a> {
                         }
                         Some(Range {
                             start: Position::new(
+                                // FIXME: Should this be `leading_node`?
```
arendjr, line 388: I'd like to call out this FIXME, since I'm not sure if there's a bug or not. If it should indeed be `leading_node`, it would make it easy to extract a common helper so I can reduce duplication with the next method.

arendjr: I'm thinking it might be an option to drop `Binding::position()` altogether and rely on `CodeRange` within `ResolvedPattern` only.

In `regex.rs` I see a comment saying:
```
// this moves the byte-range out of sync with
// the row-col range, maybe we should just
// have a Range<usize> for String bindings?
```

I think it would simplify things, although the downside is that rows/columns may need to be calculated on-demand when we need them. So I'm not sure if it will pay off.

ilevyor: what does leading_node refer to here? I don't see it anywhere in the code base. Calculating rows/columns on the fly is expensive, we do it somewhere, but best to be avoided if possible

ilevyor (approved):

crates/core/src/binding.rs
arendjr: Sorry, I meant `leading_comment`. If you look at the constructed Range, `trailing_comment` is used for both the `Position` and `end_byte`. But for the starting byte there’s a difference with the `Position`.

ilevyor: looks like a bug to me since, the byte range and the row/col mismatch

ilevyor: should also be using the is_comment utility in language/language.rs

PR #204, opened 2024-04-11
fix: consistent range start

arendjr:
Resolves the FIXME in the previous PR and trims some duplication.

@ilevyor I left the `comment.node.kind() == "comment"` checks as is. Looks like the `is_comment()` util requires a `Language` reference, which we don't have here. That was probably the reason it was already like that, even if a little hacky...

## Summary by CodeRabbit

- **Refactor**
    - Improved the handling of list variants by delegating range extraction to a new function for better code organization.


ilevyor (approved):
Title needs to conform to standard

ilevyor (approved):

PR #222, opened 2024-04-13
refactor: introduce `QueryContext` trait

arendjr:
I initially tried to introduce additional traits from the bottom up, similar to how I had done for `Context`. Unfortunately, when I tried to move onto `Binding` I got stuck. This is because some places have to remain hard-coded to `NodeWithSource` or `MarzanoContext` for now, while others had to become abstract, and compile issues popped up at the edges.

So I bit the bullet, and switched to a top-down approach: I added a generic to `Pattern` itself. I notice I even resolved a `TODO` in the code, so I think the idea is quite uncontroversial. It was kinda unavoidable it would become another big PR though :)


crates/core/src/pattern/or.rs
```diff
@@ -1,46 +1,51 @@
...
             for p in self.patterns.iter() {
                 // filter out pattern which cannot match because of a mismatched node type
-                if let (Some(binding_node), Pattern::ASTNode(node_pattern)) =
+                if let (Some(binding_node), Pattern::AstNode(node_pattern)) =
                     (binding_vector.last().and_then(Binding::as_node), p)
                 {
-                    if node_pattern.sort != binding_node.node.kind_id() {
+                    // Safety: This is safe as long as `MarzanoProblemContext` is the
+                    // only implementation of `ProblemContext`.
+                    if !node_pattern.matches_kind_of(unsafe { transmute(&binding_node) }) {
```
arendjr, line 48: This bears some attention. This is another edge where I couldn't convert the concrete `NodeWithSource` into an abstract `AstNode`. The proper fix would be to make `Binding` a trait, and then `ResolvedPattern` would be pulled in as either needing another generic or itself becoming a trait too, and the PR would've become a lot bigger yet. So it's my intention to fix this in the next PR instead.

crates/core/src/context.rs
```diff
@@ -10,25 +11,34 @@ use crate::{
...
 };
 use anyhow::Result;
+use grit_util::AstNode;
 use marzano_language::target_language::TargetLanguage;
 use marzano_util::analysis_logs::AnalysisLogs;
 
-pub trait Context {
-    fn pattern_definitions(&self) -> &[PatternDefinition];
+/// Contains various kinds of context about the problem being executed.
+pub trait ProblemContext: Clone + std::fmt::Debug + Sized {
```
morgante, line 19: I'm trying to phase out the "Problem" nomenclature. Let's call this `QueryContext`.

morgante (commented):
Looks reasonable and I think a generic is probably necessary.

Let's call it a "query" instead of a "problem" though.

crates/core/src/problem.rs
```diff
@@ -652,3 +654,12 @@ impl MatchRanges {
         }
     }
 }
+
+#[derive(Clone, Debug)]
+pub struct MarzanoProblemContext;
```
morgante, line 659:
```suggestion
pub struct MarzanoQueryContext;
```

morgante (changes requested):
WASM build is failing.

crates/core/src/pattern/or.rs
morgante: Thanks, definitely make this a priority.

morgante (approved):

PR #236, opened 2024-04-15
chore: simplify comment api in language crate

ilevyor:
replace `is_comment_wrapper` with `is_comment_node` as that is what it does, and has it also return true if the sort of the node is a comment simplifying it's usage.

## Summary by CodeRabbit

- **New Features**
  - Enhanced comment identification across various programming languages to improve accuracy and context sensitivity.
  - Added handling for JSX expressions in JavaScript and TypeScript modules.

- **Refactor**
  - Updated function names and refactored logic for better clarity and maintainability in comment-related functionalities.
  - Simplified and unified methods for checking comment nodes across different language implementations.

- **Bug Fixes**
  - Corrected the logic flow in the `Matcher` implementation and various language modules to ensure consistent behavior when identifying comments.

### DIFF
--- a/crates/language/src/language.rs
+++ b/crates/language/src/language.rs
@@ -246,10 +246,15 @@ pub trait Language: NodeTypes {
         &EXACT_REPLACED_VARIABLE_REGEX
     }
 
-    fn is_comment(&self, _id: SortId) -> bool;
+    // checks if the sort is the languages comment sort
+    fn is_comment_sort(&self, _id: SortId) -> bool;
 
-    fn is_comment_wrapper(&self, _node: &tree_sitter::Node) -> bool {
-        false
+    // checks if a node is a comment. distinct from the above
+    // in that sometimes a node is a comment but doesn't have
+    // a comment sort for example when parsing javascript,
+    // comments embedded in jsx dont have the comment sort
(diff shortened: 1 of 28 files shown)


crates/language/src/language.rs
```diff
@@ -246,10 +246,15 @@ pub trait Language: NodeTypes {
...
-
-    fn is_comment_wrapper(&self, _node: &tree_sitter::Node) -> bool {
-        false
+    // checks if the sort is the languages comment sort
+    fn is_comment_sort(&self, _id: SortId) -> bool;
+
+    // checks if a node is a comment. distinct from the above
+    // in that sometimes a node is a comment but doesn't have
+    // a comment sort for example when parsing javascript,
+    // comments embedded in jsx dont have the comment sort
```
morgante, line 255: I think we realistically want to get rid of `is_comment_sort`, since all uses of it we'd also want to be finding JSX comment nodes too.

morgante (approved):

crates/language/src/language.rs
ilevyor: No when compiling we don't care if the node is a comment we just want the sort.

PR #244, opened 2024-04-16
fix: only do openssl tricks on non-windows

Gankra:
This fixes the windows-msvc target, at least locally. However tests now fail due to snapshotting extensively using un-normalized paths, so `./` becomes `.\`, which can be fixed in several places that I don't want to make the call on.

(I also reflexively ran `cargo fmt`, lmk if you want me to revert those driveby changes)

## Summary by CodeRabbit

- **Refactor**
	- Improved portability and clarified usage of OpenSSL across different platforms.
- **Style**
	- Code cleanup including removal of unnecessary whitespace and empty lines.
- **Tests**
	- Minor test modification for enhanced functionality.


morgante:
> However tests now fail due to snapshotting extensively using un-normalized paths, so ./ becomes .\, which can be fixed in several places that I don't want to make the call on.

I'm assuming this might just be locally on Windows - let's see how CI responds.

Gankra:
Specifically:

```
  crates/cli_bin/tests/apply.rs (basic_terraform_apply)
  crates/cli_bin/tests/apply.rs (compact_output)
  crates/cli_bin/tests/apply.rs (markdown_pattern_with_invalid_name)      
  crates/cli_bin/tests/apply.rs (multifile_terraform_apply)
  crates/cli_bin/tests/apply.rs (output_jsonl)
  crates/cli_bin/tests/apply.rs (terraform_complex)
```

Although terraform_complex has an actual output change, there 4 matches becomes 0 matches.

morgante (approved):

morgante:
Opened https://github.com/getgrit/gritql/issues/245 to track the path normalization. We probably do want to run Windows tests once we have a Windows release.

PR #249, opened 2024-04-17
refactor: use traits for `Binding` and `ResolvedPatterns`

arendjr:
Sorry for another big one, it's hard to constrain myself when I'm getting this close :D Let me know if you prefer me to cut out another chunk first.

There might be a few small offenders left, but I think this pretty much clears the way for extracting the `grit-core-patterns` crate out of `marzano-core`.

A few minor structs were made abstract besides `Binding` and `ResolvedPatterns`: `File`, `AstLeafNode` and `CodeSnippet`.


crates/core/src/constant.rs
```diff
@@ -0,0 +1,50 @@
+use std::fmt::Display;
```
arendjr, line 1: This file has simply been extracted out of `binding.rs`, so that one only contains the trait now.

crates/core/src/equivalence.rs
```diff
@@ -1,64 +1,7 @@
-use crate::binding::Binding;
 use grit_util::AstNode;
 use itertools::{EitherOrBoth, Itertools};
-use marzano_language::language::Language;
 use marzano_util::node_with_source::NodeWithSource;
 
-impl<'a> Binding<'a> {
```
arendjr, line 7: Unfortunately, a few methods like this had to be moved. This is because while `impl Type` blocks can be split into multiple blocks, `impl Trait for Type` blocks cannot.

crates/core/src/pattern/or.rs
```diff
@@ -42,9 +41,7 @@ impl<Q: QueryContext> Matcher<Q> for Or<Q> {
                 // filter out pattern which cannot match because of a mismatched node type
                 if let (Some(binding_node), Pattern::AstNode(node_pattern)) = (binding.as_node(), p)
                 {
-                    // Safety: This is safe as long as `MarzanoProblemContext` is the
-                    // only implementation of `ProblemContext`.
-                    if !node_pattern.matches_kind_of(unsafe { transmute(&binding_node) }) {
```
arendjr, line 47: The `unsafe` is gone again :partying_face:

crates/core/src/context.rs
```diff
@@ -16,10 +18,15 @@ use marzano_language::target_language::TargetLanguage;
 use marzano_util::analysis_logs::AnalysisLogs;
 
 /// Contains various kinds of context about the query being executed.
-pub trait QueryContext: Clone + std::fmt::Debug + Sized {
+pub trait QueryContext: Clone + std::fmt::Debug + Sized + 'static {
```
ilevyor, line 21: Why do we need to add the static lifetime here? I'm a little nervous it might make QueryContext overly inflexible

crates/core/src/test.rs
```diff
@@ -12833,7 +12833,7 @@ fn rust_match_fn_params() {
                 |impl Matcher for FilePattern {
                 |    fn execute<'a>(
                 |        &'a self,
-                |        resolved_pattern: &ResolvedPattern<'a>,
+                |        resolved_pattern: &Q::ResolvedPattern<'a>,
```
ilevyor, line 12836: big find and replace casualty?  Change is fine, just curious why.

ilevyor (approved):

crates/core/src/test.rs
arendjr: Ah, probably:)

PR #257, opened 2024-04-19
chore: move utils to `grit-util`

arendjr:
This moves a few utils to `grit-utils`, making them free of TreeSitter dependencies in the process.

Namely:
- `Point`, `Range` and associated structs. These were tied to TreeSitter through their `From<T>` implementations, but this was covered by consistently doing the conversion inside `NodeWithSource::range()` instead. Some minor helper functions were cleaned up as well.
- `AnalysisLogs`, since it has no other dependencies than `Point` and `Range`.

I have also cleaned up the `resolve!` and `resolve_opt!` macros, since it appeared almost all of their usages were gone already. I also ran into `linearize.rs`, which appeared to be fully unused, so I removed it as well.


crates/core/src/pattern_compiler/snippet_compiler.rs
```diff
@@ -105,7 +102,7 @@ pub(crate) fn dynamic_snippet_from_source(
         parts.push(DynamicSnippetPart::String(
             source[last as usize..byte_range.start as usize].to_string(),
         ));
-        let start_pos = Position::from_byte_index(source, Some((last_pos, last)), byte_range.start);
+        let start_pos = position_from_previous(last_pos, source, last, byte_range.start);
```
arendjr, line 105: I removed the middle argument from `Position::from_byte_index()`, because it was only used for this specific call site, and made the function a bit confusing. The functionality was replaced through this `position_from_previous()` function which directly wraps `Position::from_byte_index()`.

crates/grit-util/src/position.rs
```diff
@@ -1,40 +1,63 @@
...
+        for line in source[..byte_index].lines() {
+            line_count += 1;
+            last_line = line;
+        }
+        let last_line_length = last_line.len() as u32;
+        Self::new(line_count, last_line_length + 1)
+    }
+
+    /// Returns the byte index for this `Position` within the given `source`.
+    pub fn byte_index(&self, source: &str) -> usize {
```
arendjr, line 49: This used to be the `get_one_indexed_position_offset()` function, but I've reimplemented it so it doesn't allocate anymore and can exit early if the position is early in the file.

crates/grit-util/src/position.rs
```diff
@@ -149,73 +95,11 @@ pub struct Range {
...
-    node.end_byte()
-}
-
-#[cfg(target_arch = "wasm32")]
-pub fn end_byte(node: &Node, text: &str) -> u32 {
-    char_index_to_byte_index(node.end_byte(), text)
-}
-
 impl Range {
-    pub fn char_range_to_byte_range(self, context: &str) -> Self {
```
arendjr, line 186: This function turned out to be used in tests only.

arendjr:
>  I'm not sure we want to ditch resolve.

Happy to put it back, but I don't think it's considered idiomatic since we have `if`...`else` syntax.

morgante:
> Happy to put it back, but I don't think it's considered idiomatic since we have if...else syntax.

Ah yeah, it's not doing much for us. Let's remove it.

ilevyor (approved):

Issue #250, opened 2024-04-17
Python indentation issue - error


jonathan-s:
This is in fact valid python code. Though it throws an error when trying to do something with it. Check it here https://extendsclass.com/python-tester.html or in your local environment. 

```python
def test_yaml_file():
    """some test comment"""
    variable = """
en:
  formtastic:
    labels:
      title: "Title"  # Default global value
      article:
        body: "Article content"
      post:
        new:
          title: "Choose a title..."
          body: "Write something..."
        edit:
          title: "Edit title"
          body: "Edit body"
        """
```

Also worth noting that the indentation inside the comment IS important in this case, so it should not be re-formatted.

ilevyor:
We assume that a binding corresponds to a scope, and therefore preserves indentation. multiline strings can violate this assumption, I'll add some special casing logic for this

PR #259, opened 2024-04-19
fix: don't pad python strings

ilevyor:
fixes: https://github.com/getgrit/gritql/issues/250
skips padding for python strings.

## Summary by CodeRabbit

- **New Features**
	- Enhanced snippet handling and padding logic in the core processing module.
	- Introduced new methods for node identification and type retrieval in utility modules.
	- Improved pattern matching and effect application in the core pattern module.
	- Added support for handling multiline strings in Python test functions.

- **Refactor**
	- Updated function calls and parameters to improve clarity and maintainability across various modules.

- **Tests**
	- Added new test functions to verify the handling of multiline strings in Python.

- **Documentation**
	- No visible changes to end-users.


crates/language/src/language.rs
```diff
@@ -257,6 +257,10 @@ pub trait Language: NodeTypes {
         self.is_comment_sort(node.node.kind_id())
     }
 
+    fn skip_padding_sort(&self, _id: SortId) -> bool {
```
morgante, line 260: Add a comment explaining why we need this / what it does.

crates/core/src/marzano_binding.rs
```diff
@@ -52,14 +53,35 @@ impl PartialEq for MarzanoBinding<'_> {
...
+}
+
+pub(crate) fn pad_snippet(padding: &str, snippet: &str, lang: &impl Language) -> Result<String> {
     let mut lines = snippet.split('\n');
     let mut result = lines.next().unwrap_or_default().to_string();
 
     // Add the rest of lines in the snippet with padding
-    lines.for_each(|line| result.push_str(&format!("\n{}{}", &padding, line)));
-    result
+    let skip_ranges = get_skip_padding_ranges_for_snippet(lang, snippet)?;
```
morgante, line 75: This doesn't look optimal, why are we needing to re-parse here?

crates/core/src/text_unparser.rs
```diff
@@ -34,16 +35,16 @@ pub(crate) fn apply_effects<'a, Q: QueryContext>(
...
     }
     let mut memo: HashMap<CodeRange, Option<String>> = HashMap::new();
     let (from_inline, ranges) = linearize_binding(
         language,
         &effects,
         files,
         &mut memo,
-        code,
-        CodeRange::new(0, code.len() as u32, code),
+        code.clone(),
```
morgante, line 46: More clones isn't good.

crates/core/src/marzano_binding.rs
```diff
@@ -73,14 +95,27 @@ fn adjust_ranges(substitutions: &mut [(EffectRange, String)], index: usize, delt
     }
 }
 
+pub(crate) fn is_index_in_range(index: u32, skip_ranges: &[CodeRange]) -> bool {
```
morgante, line 98:
```suggestion
pub(crate) fn is_index_in_ranges(index: u32, skip_ranges: &[CodeRange]) -> bool {
```

morgante (changes requested):

crates/core/src/marzano_binding.rs
ilevyor: we're only parsing a snippet, it's to check if it's a string, if it's a string we don't want to pad it

morgante: Sure, but we don't actually know/guarantee RHS snippets are parseable. We had the original binding at some point, we should be able to get the sort from there.

ilevyor: The original binding if there was one is lost at this point as a snippet is made up of a patchwork of bindings that have been linearized by this point. It doesn't matter if we the snippet is parsed without error, all we are looking for are strings if we get error nodes instead of strings then the worst case is we padded when we should have, but I find it unlikely that we will have a node with a string that doesn't parse as a string node. Consider the rhs `"""$content"""` content wasn't a string, but we turned it into a string on the rhs, so there's no original binding with the information we need.

morgante: Ok, I still don't love this approach but it's probably only a minor performance hit.

crates/core/src/marzano_binding.rs
```diff
@@ -73,14 +95,27 @@
...
+    skip_ranges
+        .iter()
+        .any(|r| r.start <= index && index < r.end)
+}
+
+// safety ensure that sup and sub are slices of the same string.
+pub(crate) fn get_slice_byte_offset(sup: &str, sub: &str) -> u32 {
+    unsafe { sub.as_ptr().byte_offset_from(sup.as_ptr()) }.unsigned_abs() as u32
+}
```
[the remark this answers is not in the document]
morgante: I'd prefer to avoid adding unsafe code for a minor fix. If this would be safer inline, could we just do it there?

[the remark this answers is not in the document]
ilevyor: This fix is much slower, and introduces a bug to the code. If the substring appears more than once in the superstring it cannot differentiate between between the the index of out substring, and another substring with the same content.

morgante (approved):

PR #266, opened 2024-04-22
refactor: split `Language` trait

arendjr:
Another big one. This splits the `Language` trait into `Language` and `MarzanoLanguage`, with the former being TreeSitter-agnostic, while the latter still maintains TreeSitter-specific logic.

There is now also a TreeSitter-agnostic `Parser` trait that's used to give language implementations access to file and snippet parsing.


crates/cli/src/result_formatting.rs
```diff
@@ -76,7 +76,7 @@ fn print_all_done(item: &AllDone, f: &mut fmt::Formatter<'_>) -> fmt::Result {
 }
 
 fn print_error_log(log: &AnalysisLog, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-    let file_prefix = if log.file == DEFAULT_FILE_NAME {
+    let file_prefix = if log.file.is_empty() || log.file == DEFAULT_FILE_NAME {
```
arendjr, line 79: The new `Parser` trait accepts an optional `Path` for the filename, so it doesn't need to depend on the `DEFAULT_FILE_NAME` constant. When no path is given, this will end up as an empty filename in the logs.

crates/core/src/analysis.rs
```diff
@@ -1,22 +1,21 @@
...
-use tree_sitter::Parser;
+use std::path::Path;
 
 /// Walks the call tree and returns true if the predicate is true for any node.
 /// This is potentially error-prone, so not entirely recommended
 fn walk_call_tree(
     node: &NodeWithSource,
     libs: &BTreeMap<String, String>,
-    grit_parser: &mut Parser,
+    grit_parser: &mut MarzanoGritParser,
```
arendjr, line 14: Introducing the `MarzanoGritParser` wasn't strictly necessary, because in the end it doesn't use the `Parser` trait either. But it does return the same new `Tree` data structure, so it helps make things a bit more consistent.

crates/core/src/marzano_binding.rs
```diff
@@ -53,26 +51,12 @@ impl PartialEq for MarzanoBinding<'_> {
...
-) -> Result<Vec<CodeRange>> {
-    let mut parser = TSParser::new()?;
-    parser.set_language(lang.get_ts_language())?;
-    let tree = parser
-        .parse(snippet, None)?
-        .ok_or(anyhow!("failed to parse snippet"))?;
-    let root = tree.root_node();
-    let node = NodeWithSource::new(root, snippet);
-    Ok(get_skip_padding_ranges(&node, lang))
-}
```
arendjr, line 68: Moved onto the `Language` trait.

crates/core/src/marzano_context.rs
```diff
@@ -0,0 +1,273 @@
...
+    // FIXME: Don't depend on Grit's file handling in context.
+    fn files(&self) -> &FileOwners {
+        self.files
+    }
+
+    fn language(&self) -> &TargetLanguage {
+        self.language
+    }
+
+    fn exec_step(
```
arendjr, line 121: This trait implementation grew significantly due to the implementatoin of this `exec_step()` method, which was moved from the `Step` pattern. Since I was moving anyway, I put the entire `MarzanoContext` into its own file.

Note how the method above used to have a FIXME, which is now resolved :)

crates/core/src/parse.rs
```diff
@@ -1,56 +1,43 @@
...
-#[cfg(feature = "grit-parser")]
-pub fn make_grit_parser() -> Result<Parser> {
-    use anyhow::Context;
-
-    let mut parser = Parser::new().unwrap();
-    parser
-        .set_language(&tree_sitter_gritql::language().into())
-        .with_context(|| "Failed to load grit language")?;
-    Ok(parser)
-}
```
arendjr, line 47: This is replaced with the `MarzanoGritParser` in the `marzano-language` crate.

crates/core/src/pattern_compiler/compiler.rs
```diff
@@ -78,58 +77,6 @@ pub(crate) struct NodeCompilationContext<'a> {
...
-            let log = log_builder
-                .clone()
-                .message(message)
-                .position(position)
-                .build()?;
-            errors.push(log);
-        }
-    }
-    Ok(errors.into())
-}
```
arendjr, line 131: Moved to the parser.

crates/grit-util/src/language.rs
```diff
@@ -0,0 +1,155 @@
+use crate::{constants::*, traverse, AstNode, CodeRange, Order, Range};
+use regex::Regex;
+
+pub enum GritMetaValue {
+    Underscore,
+    Dots,
+    Variable(String),
+}
+
+pub trait Language: Sized {
```
arendjr, line 10: Most of `Language` was fortunately TreeSitter-agnostic from the start.

crates/language/src/csharp.rs
```diff
@@ -63,11 +62,25 @@ impl Language for CSharp {
...
 
-    fn metavariable_sort(&self) -> SortId {
-        self.metavariable_sort
+    fn is_comment(&self, node: &NodeWithSource) -> bool {
+        MarzanoLanguage::is_comment_node(self, node)
+    }
+
+    fn is_metavariable(&self, node: &NodeWithSource) -> bool {
+        MarzanoLanguage::is_metavariable_node(self, node)
+    }
```
arendjr, line 71: These are unfortunately reoccurring. Code that is otherwise TreeSitter-agnostic wants to know whether something is a comment or a metavariable, but the implementations for that are tied to TreeSitter, so we end up forwarding from one trait to the other.

crates/util/src/node_with_source.rs
```diff
@@ -143,6 +143,10 @@ impl<'a> AstNode for NodeWithSource<'a> {
         CodeRange::new(self.node.start_byte(), self.node.end_byte(), self.source)
     }
 
+    fn full_source(&self) -> &str {
+        self.source
+    }
```
arendjr, line 148: This method was only necessary after fixing the last merge conflict. It is what it is :)

crates/language/src/language.rs
```diff
@@ -170,156 +136,149 @@ pub trait NodeTypes {
...
-    /// tree sitter language to parse the source
-    fn get_ts_language(&self) -> &TSLanguage;
-
-    fn language_name(&self) -> &'static str;
-
-    fn skip_snippet_compilation_of_field(&self, _sort_id: SortId, _field_id: FieldId) -> bool {
-        false
-    }
+#[derive(Clone, Debug)]
+pub struct Tree {
```
arendjr, line 140: The `Tree` data structure encapsulates a TreeSitter tree, while implementing the `Ast` trait. That one just has a single method to get the root node as an `AstNode`.

crates/language/src/language.rs
```diff
@@ -392,116 +308,88 @@ fn file_parsing_error(
...
-    pub snippet_postfix: &'static str,
-    pub snippet: String,
-    pub context: String,
-    pub snippet_start: u32,
-    pub snippet_end: u32,
-    _metavariable_sort: SortId,
-}
-
-#[derive(Debug, Clone)]
-pub struct SnippetNode<'a> {
```
arendjr, line 422: The `SnippetNode` struct has been completely replaced with `NodeWithContainer`.

crates/core/src/context.rs
```diff
@@ -64,8 +64,15 @@ pub trait ExecContext<Q: QueryContext> {
...
-    fn language(&self) -> &impl Language;
+    fn language(&self) -> &Q::Language<'a>;
+
+    fn exec_step(
+        &'a self,
+        step: &'a Pattern<Q>,
+        binding: &Q::ResolvedPattern<'a>,
+        state: &mut State<'a, Q>,
+        logs: &mut AnalysisLogs,
+    ) -> Result<bool>;
```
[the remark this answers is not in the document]
ilevyor: That's fine, but note that *all* grit queries use step. It essentially generates the tree that we process, and applies any rewrites

crates/cli_bin/tests/snapshots/plumbing__checks_invalid_patterns.snap
```diff
@@ -1,10 +1,9 @@
 ---
-source: apps/marzano/cli/tests/plumbing.rs
+source: crates/cli_bin/tests/plumbing.rs
 expression: "String::from_utf8(result.stderr)?"
 ---
 Error: Failed to compile pattern `console.lo: engine_id: Some("marzano(0.1)")
-file_name: Some("PlaygroundPattern")
+file_name: None
```
morgante, line 6: @arendjr @ilevyor This should *NOT* have been merged. Breaking this snapshot test causes lots of downstream problems.

PR #273, opened 2024-04-24
Revert "refactor: split `Language` trait"

morgante:
Reverts getgrit/gritql#266

This broke snapshot tests, which is a pretty big indicator that it broke downstream compatibility (indeed, it broke our web app).

@arendjr Please make sure you don't break any tests. Pure refactoring should not change the behavior of Grit. If you want to change the interface, please do it in a separate (small) PR, not as part of a massive refactor.


arendjr:
My bad, I misjudged the possible impact of that value. I'll recreate the PR while trying to stick to the old semantics with the "PlaygroundPattern".

After that I'll make another one trying to get rid of the "PlaygroundPattern" constant altogether, because it does seem like a clear case of technical debt if every low-level crate needs to account for a special string value that only affects the playground. But there won't be any rush for that one, so you can adapt the web app when you feel like so.

PR #275, opened 2024-04-24
feat: fetch rules

seren5240:
fixes https://github.com/getgrit/rewriter/issues/9443

I didn't want to add another dependency because we already have too many, maybe we can add one for GraphQL if we end up using more of it.

## Summary by CodeRabbit

- **New Features**
	- Integrated a new GraphQL API for fetching auto review rules for projects.
	- Added handling for new environment variables to enhance configuration management.

- **Refactor**
	- Updated environment variable usage across various modules to use constants, improving code maintainability and consistency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1924,6 +1924,7 @@ dependencies = [
  "log",
  "marzano-auth",
  "marzano-core",
+ "marzano-graphql",
  "marzano-gritmodule",
  "marzano-language",
  "marzano-lsp",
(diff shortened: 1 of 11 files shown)


morgante (changes requested):
I think we should have a separate crate for calling the API.

seren5240:
> I think we should have a separate crate for calling the API.

It's relying on the updater. How do you want to handle this?

morgante:
> > I think we should have a separate crate for calling the API.
> 
> It's relying on the updater. How do you want to handle this?

Use [auth](https://github.com/getgrit/gritql/pull/275/files#diff-5cbe23550009fba774c7958956af6a36df3388345c4dbff26446fad5316cf305R68) as an argument here, instead of calling it in the function.

morgante (approved):

PR #279, opened 2024-04-25
chore: rename `grit-core-patterns` to `grit-pattern-matcher`

arendjr:
Also added a `README.md` and a crate description.

I've made two other minor changes to the crate:

* The `pattern` module now directly re-exports all the inner patterns, instead of exposing their modules. This makes importing them slightly more convenient for consumers. And since the imports needed to be updated anyway, I figured the impact on this PR was minor.
* ~~The `grit-util` crate is re-exported under `grit_pattern_matcher::util`, again to ease usage by consumers. In theory I think we could even roll the util crate into the pattern matcher, but I think it is fine as is.~~

## Summary by CodeRabbit

- **New Features**
  - Renamed the internal pattern matching module to enhance clarity and consistency across the platform.
  - Improved pattern matching capabilities with updated module references, ensuring smoother functionality and integration.
- **Bug Fixes**
  - Fixed issues related to pattern imports and usage across various components, enhancing stability and performance.
- **Refactor**
  - Extensive updates to import paths and module usage to align with the new pattern matching module name, improving code maintainability and future scalability.
- **Chores**
  - Updated dependencies and internal references to support the new module structure.


crates/core/src/constants.rs
```diff
@@ -1 +1,3 @@
-pub use grit_core_patterns::constants::DEFAULT_FILE_NAME;
+pub use grit_pattern_matcher::constants::DEFAULT_FILE_NAME;
+
+pub const MAX_FILE_SIZE: usize = 1_000_000;
```
arendjr, line 3: This constant was moved from `grit-pattern-matcher`'s constants, since it's only used by `marzano-core`.

crates/grit-pattern-matcher/src/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+pub mod constant;
+pub mod constants;
+pub mod context;
+pub mod effects;
+pub mod errors;
+pub mod file_owners;
+pub mod intervals;
+pub mod pattern;
+
+pub use grit_util as util;
```
morgante, line 17: @arendjr Let's not do this. I generally prefer to avoid exporting/aliasing crates - consumers should be fine with importing grit_util directly as well.

crates/grit-pattern-matcher/README.md
```diff
@@ -0,0 +1,9 @@
+# `grit-pattern-matcher`
+
+This crate contains all the pattern definitions that are at the heart of the
+GritQL engine. There's a `Matcher` trait that's implemented by the patterns,
+which implements the matching logic.
+
+It is important this crate stays free of TreeSitter dependencies, since it is
```
morgante, line 7: It would be a good opportunity to add a Grit rule to enforce that actually, in `.grit`, otherwise I'm sure someone will accidentally try. 

Even something basic like `tree_sitter` should be sufficient.

morgante (changes requested):
Thanks for handling this.

crates/grit-pattern-matcher/src/lib.rs
arendjr: I'll remove it.

I think the main reason why it's easier for consumers is when they need to upgrade. If they use the dependent crate through a re-export, they can treat it as a transitive dependency. So it will be upgraded along with the direct dependency, which will also take care of making sure the version they're using is a compatible one. If they import `grit_util` directly, they are responsible themselves for upgrading it along with `grit_pattern_matcher`. It's usually not a big deal though, since mistakes tend to be pointed out as compile errors :)

morgante: Yeah, there are pros and cons but since it has utils that consumers will directly use I think they should import it directly.

crates/grit-pattern-matcher/README.md
arendjr: Done. Do I need to add anything to make sure CI runs it, or does it always check all patterns in `grit.yaml`?

morgante: That's all you need to do, grit check picks it up automatically.

ilevyor (approved):

PR #292, opened 2024-04-30
refactor!: use `ByteRange` internally instead of `Range`

arendjr:
Internally, Grit uses the `Range` type which tracks both byte indices as well as line and column numbers. This is relatively easy to implement on top of TreeSitter, since the TS parser provides line and column numbers. Unfortunately, the Biome parser doesn't.

Since I wondered about the usage of the `Range` type before, I gave it a shot to see how hard it is to replace the internal usages of `Range` with `ByteRange`. Seems that barring a handful of cases, this was pretty doable. The main outlier was `Pattern::Range`, for which I had to keep a few `Range` usages around.

**Question:** What is `Pattern::Range` actually used for? I haven't seen it documented anywhere. If it's a legacy thing, it would be great if it could be cleaned up.

Of course there is a danger of performance degradation, because in a few places we now need to calculate the line and column numbers where previously they were simply handed to us. But on the upside, those places are now all on the "outside" of the engine (such as the `api` module which I've kept the same to minimize the risk of disruption), while we can actually avoid some on-the-spot calculations inside the engine, such as for metavariable ranges. So while I don't have a good benchmark at my disposal, I suspect the performance impact is negligible. It might even be slightly positive, since there are quite some code simplifications across the board as well, and the data types have become slightly smaller too.

All that said, this is definitely a risky PR, so please verify it doesn't break anything on your end.

## Summary by CodeRabbit

- **Refactor**
  - Simplified field names and code for consistency and clarity in handling ranges.
  - Modified function signatures and mappings to enhance precision in byte range handling.
  
- **New Features**
  - Enhanced range initialization in the `Match` struct to improve accuracy in source mapping.
  
- **Bug Fixes**
  - Adjusted newline handling at the end of files to prevent formatting errors.

- **Documentation**
  - Updated versioning and authorship details to reflect the current project status.

- **Style**
  - Standardized usage of `ByteRange` over `Range` across various modules to ensure uniformity in code.


arendjr:
The diff for [apply__output_jsonl.snap](https://github.com/getgrit/gritql/pull/292/files#diff-ddb9f0ad373ae261019c2d6103cd3c7c43eebbc342866af03280938d34baa192) doesn't show on GitHub, and it's a rough one to inspect. But if appears the only things that changed is the omission of line and column numbers, similar to other changed snapshots.

crates/core/src/pattern_compiler/snippet_compiler.rs
```diff
@@ -94,20 +94,14 @@ pub(crate) fn dynamic_snippet_from_source(
...
+            source[last..byte_range.start].to_string(),
         ));
-        let start_pos = position_from_previous(last_pos, source, last, byte_range.start);
-        // todo: does this handle utf8 correctly?
-        last_pos = Position::new(start_pos.line, start_pos.column + var.len() as u32);
-        let range = Range::new(
-            start_pos,
-            last_pos,
-            source_range.start_byte + byte_range.start,
-            source_range.start_byte + byte_range.start + var.len() as u32,
```
arendjr, line 110: The snippet compiler has been simplified since it doesn't need to scan the source anymore to construct new `Position`s.

crates/grit-pattern-matcher/src/pattern/regex.rs
```diff
@@ -93,16 +93,13 @@ impl<Q: QueryContext> RegexPattern<Q> {
...
                     .get_last_binding()
-                    .map(|binding| (binding.position(context.language()), binding.source()))
+                    .map(|binding| (binding.range(context.language()), binding.source()))
                 {
-                    // this moves the byte-range out of sync with
-                    // the row-col range, maybe we should just
-                    // have a Range<usize> for String bindings?
-                    position.end_byte = position.start_byte + range.end as u32;
-                    position.start_byte += range.start as u32;
-                    Q::ResolvedPattern::from_range_binding(position, source)
```
arendjr, line 105: Here is a gotcha from the previous implementation that won't be a concern anymore.

crates/grit-pattern-matcher/src/pattern/state.rs
```diff
@@ -267,4 +263,4 @@ impl<'a, Q: QueryContext> State<'a, Q> {
     }
 }
 
-type VarRegistry<'a, P> = Vector<Vector<Vector<Box<VariableContent<'a, P>>>>>;
+pub type VarRegistry<'a, P> = Vector<Vector<Vector<Box<VariableContent<'a, P>>>>>;
```
arendjr, line 266: I made this one public so Biome can import it too.

crates/grit-util/src/position.rs
```diff
@@ -35,16 +35,27 @@ impl Position {
...
     pub fn from_byte_index(source: &str, byte_index: usize) -> Self {
-        let mut pos = Self::first();
-        for c in source[..byte_index].chars() {
+        Self::from_relative_byte_index(Self::first(), 0, source, byte_index)
+    }
+
+    /// Create a position for the given `byte_index` in the given `source`,
+    /// counting from the given other position. This avoids double work in case
+    /// one position lies after another.
+    pub fn from_relative_byte_index(
```
arendjr, line 44: This helps to avoid overhead when creating a `Range` from a `ByteRange`.

morgante:
This is definitely the direction I've wanted to go in, so generally supportive of this refactor!

We will need to make sure the interfaces don't change so that [this](https://github.com/getgrit/gritql/blob/6a4428c3f482df0e2bc148b28da039575e734ed4/crates/cli_bin/tests/apply.rs#L1345) still works.

> Question: What is Pattern::Range actually used for? I haven't seen it documented anywhere. If it's a legacy thing, it would be great if it could be cleaned up.

It's not so much legacy as hidden. In this repo, we filter ranges as part of `--only-in-json` and `--only-in-diff`. We also have some use cases within our SaaS where we want to explicitly target certain ranges.

crates/cli_bin/tests/snapshots/parse__correct_variable_ranges_in_snippet_with_multiple_contexts.snap
```diff
@@ -20,18 +20,11 @@ variables:
...
     ranges:
-      - start:
-          line: 1
-          column: 16
-        end:
-          line: 1
-          column: 21
-        startByte: 15
-        endByte: 20
+      - start: 15
```
morgante, line 23: I thought you were leaving the external API the same? We do need to make sure these ranges aren't changed, because they're used to highlight metavariables in the Studio and Monaco expects lines/columns.

crates/cli_bin/tests/parse.rs
```diff
@@ -353,27 +353,15 @@ fn no_extraneous_ranges_for_multiple_metavariables_in_snippet_with_rewrite() ->
...
-    let lhs_end = lhs.get("end").unwrap();
-    let lhs_end_line = lhs_end.get("line").unwrap().as_u64().unwrap();
-    let lhs_end_column = lhs_end.get("column").unwrap().as_u64().unwrap();
-    assert_eq!(lhs_start_line, 1);
-    assert_eq!(lhs_start_column, 26);
-    assert_eq!(lhs_end_line, 1);
-    assert_eq!(lhs_end_column, 38);
+    let lhs_start = lhs.get("start").unwrap().as_u64().unwrap();
+    let lhs_end = lhs.get("end").unwrap().as_u64().unwrap();
+    assert_eq!(lhs_start, 25);
```
morgante, line 358: These shouldn't change.

crates/core/src/snapshots/marzano_core__test__add_to_set.snap
```diff
@@ -27,14 +27,8 @@ expression: results
...
-          - start:
-              line: 5
-              column: 1
-            end:
-              line: 5
-              column: 4
-            startByte: 48
-            endByte: 51
+          - start: 48
+            end: 51
```
morgante, line 31: (Same note as before for all tests).

arendjr:
> We will need to make sure the interfaces don't change so that [this](https://github.com/getgrit/gritql/blob/6a4428c3f482df0e2bc148b28da039575e734ed4/crates/cli_bin/tests/apply.rs#L1345) still works.

The test still passes, so I think it works :)

> I think it should be possible to compute just-in-time though - if we're looking at a file and you encounter a `Pattern::Range` only then use the offsets to retrieve the col and line.

Right now `Pattern::Range` still works because I left `ResolvedPattern::position()` intact which still returns the old `Range` with `Position` fields. Supporting that method might be a little trickier on the Biome side, but as long as I have a `source` reference in the resolved patterns there it'll be fine. For you the current approach should be fine too, since `MarzanoResolvedPattern` can even still pull the line/col information from TS. The only time it'll compute just-in-time is for a `Pattern::Range` that resolves to a `MarzanoResolvedPattern::String`.

And of course in the `api` module there are now some just-in-time computes being done.

crates/grit-util/src/ranges.rs
```diff
@@ -245,11 +270,11 @@ impl MatchRanges {
 pub struct VariableMatch {
     pub name: String,
     pub scoped_name: String,
-    pub ranges: Vec<Range>,
+    pub ranges: Vec<ByteRange>,
```
morgante, line 273: This is part of the API, so it will need to be a Range (with lines and columns).

morgante (changes requested):
Overall looks good, I think the main change we need is just to preserve Range in the variable match outputs.

crates/grit-util/src/ranges.rs
arendjr: Ah, that's why the snapshots changed :) I knew something leaked through, I just hadn't spotted where. I guess I'll just give the `api` module its own `ApiVariableMatch` type uses `Range` still. Because this type is used internally as well, and that would trigger the on-demand computations at various places inside.

arendjr: This is going to take a bit more work. In `api.rs: Match::file_to_match` we have the matches, but we no longer have access to the source of the patterns in which the variable definitions exist...

morgante: That's for the pattern file right? Realistically you might need to attach those at some point earlier in the process (ex. right after compilation).

arendjr: Think I got it fixed.

morgante: Yep, snapshots look clean to me. Thanks!

crates/grit-util/src/ranges.rs
```diff
@@ -189,13 +178,40 @@ impl From<RangeWithoutByte> for UtilRange {
...
+    pub fn new(start: usize, end: usize) -> Self {
+        Self { start, end }
+    }
+
+    pub fn abbreviated_debug(&self) -> String {
+        format!("[{}-{}]", self.start, self.end)
+    }
+
+    /// Converts a range expressed in byte indices to a range expressed in
+    /// character offets.
```
morgante, line 198:
```suggestion
    /// character offsets.
```

morgante (approved):

arendjr:
@morgante To get these changes published, do you prefer me to bump the version as part of the PR, or we use separate PRs for that?

morgante:
> @morgante To get these changes published, do you prefer me to bump the version as part of the PR, or we use separate PRs for that?

Bumping in this PR is fine.

arendjr:
Bumped.

PR #293, opened 2024-04-30
fix(updater): cross-platform compatibility, robustness enhancements

mistydemeo:
This makes a few changes to the updater with the goal of ensuring it runs across all platforms and that it requires fewer external tools. Since these tools aren't preinstalled on Windows, it should ensure that running on Windows just works the same as on other platforms. In particular:

* The external `curl` call has been replaced by a call to `reqwest`, which was already used in the crate for other things. This should also ensure it works on minimal Linux docker images, some of which come with `wget` but no `curl` preinstalled.
* The calls to Unix commands to move and rename files have been replaced with calls to Tokio's `async_fs`.
* Executable file permissions are now handled via Rust stdlib calls, and are only run on Unix; these permissions don't exist on Windows.
* The temporary paths to which tarballs are downloaded are now deleted after being unpacked.

## Summary by CodeRabbit

- **New Features**
    - Enhanced the updater tool to automatically clean up temporary files after use.
    - Improved error handling during file downloads and operations.
- **Refactor**
    - Updated file downloading, moving, and permissions setting methods for better performance and reliability.

### DIFF
--- a/crates/cli/src/updater.rs
+++ b/crates/cli/src/updater.rs
@@ -444,17 +444,14 @@ impl Updater {
     async fn download_artifact(&self, app: SupportedApp, artifact_url: String) -> Result<PathBuf> {
         let target_path = self.bin_path.join(format!("{}-temp", app.get_bin_name()));
 
-        // Download via curl
-        let cmd = AsyncCommand::new("curl")
-            .arg("-L")
-            .arg("-o")
-            .arg(&target_path)
-            .arg(&artifact_url)
-            .output()
-            .await?;
-
-        if !cmd.status.success() {
-            bail!("Failed to download artifact: {:?}", cmd);
+        match reqwest::get(&artifact_url).await {
+            Ok(response) => {
+                let contents = response.bytes().await?.to_vec();
+                async_fs::write(&target_path, contents).await?;
+            }
+            Err(e) => {
+                bail!("Failed to download artifact: {:?}", e);
+            }
         }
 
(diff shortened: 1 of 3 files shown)


crates/cli/src/updater.rs
```diff
@@ -444,17 +447,14 @@ impl Updater {
...
-            .arg(&artifact_url)
-            .output()
-            .await?;
-
-        if !cmd.status.success() {
-            bail!("Failed to download artifact: {:?}", cmd);
+        match reqwest::get(&artifact_url).await {
+            Ok(response) => {
+                let contents = response.bytes().await?.to_vec();
+                async_fs::write(&target_path, contents).await?;
```
morgante, line 453: Is there a way to stream directly into the file, so we don't have to load it all into memory?

morgante (approved):

crates/cli/src/updater.rs
mistydemeo: We can do that! Streaming support is an extra feature in reqwest; are you okay with this expanding the dependency tree a little?

morgante: That's fine, we're not very disciplined about binary size.

mistydemeo: All right, pushed a change that streams the response and writes chunks to disk.

morgante: Thanks, looks good!

morgante (approved):

Issue #51, opened 2024-03-22
Watch mode for `grit patterns test`


morgante:
We would like to have a watch mode for `grit patterns test --watch`.

Watch mode should be similar to other testing/CLI tools that offer a watch mode: when files change, re-run tests.

## Acceptance criteria
- Watch mode should watch for file changes to any patterns in the `.grit` directory and automatically re-run those tests
- If a file changes, only the *affected* patterns should be re-run. Ex. if I edit [this file](https://github.com/getgrit/stdlib/blob/main/.grit/patterns/js/add_playwright_tags.md) then only that pattern test should re-run.
- You will also need to account for pattern dependencies. Ex. if https://github.com/getgrit/stdlib/blob/main/.grit/patterns/js/imports.grit is edited then any patterns which use the `ensure_import_from` pattern should be re-run. Utilities like [this](https://github.com/getgrit/gritql/blob/c8f9160197e68aa5340a45f19005706a13a60f30/crates/core/src/pattern/compiler.rs#L563) can be leveraged for walking the call graph
- At least 2 tests should be included, including ideally at least one integration test of the final binary like we do [here](https://github.com/getgrit/gritql/tree/main/crates/cli_bin/tests).

webbdays:
@el-buku 
any progress.
still on this?

el-buku:
For sure, started earlier I’ll open a draft PR

vivasvan1:
@Sambit003 @el-buku 
Hi there fellas,

Are you working on this? 
If yes, let me know if you interested in working togather.
If no, I would like to attempt it.

Hope u have a nice day.

varshith257:
@morgante Is this issue open to grab? 
Have some good approach in my mind

morgante:
@varshith257 No, this is already complete and I will award the bounty to @Ashu999. Feel free to open a PR with an improved approach if you have one in mind.

varshith257:
@morgante It seems already PR is merged. How my work reflects it and rewarded?

morgante:
@varshith257 You did no work for that PR. Why would you be rewarded for something you made no contributions to?

varshith257:
@morgante I think we misunderstood. You said to open a PR with my approach. So that's I am asking for 😅

morgante:
@varshith257 This is an open source repository, under a permissive license. If you have a superior approach, provide a PR. We will not be paying for every PR.

If you expect every single PR on every repo will be "compensated" you do not understand the ethos of open source.

[...]

PR #298, opened 2024-05-01
feat: watch mode for patterns test

Ashu999:
Fixes: #51
/claim #51

## Summary by CodeRabbit

- **New Features**
  - Introduced watch mode for testing patterns with `grit patterns test --watch`.
  - Added YAML support for pattern files alongside Grit and Markdown formats.

- **Enhancements**
  - Improved dependency management and added new dependencies.
  - Added error handling and logging for unsupported output formats.
  - Enhanced pattern testing functionality and file modification tracking.

- **Bug Fixes**
  - Improved error handling for empty patterns in test results.

- **Testing**
  - Added new test cases for watch mode functionality.
  - Introduced integration test patterns for dependency management and code restrictions.

- **Documentation**
  - Updated `.gitignore` to ignore specific log files and grit modules.

## Greptile Summary

--
- Introduced watch mode for pattern testing with real-time file monitoring
- Enhanced error handling and logging in pattern testing
- Modified function parameters for better efficiency in `compiler.rs`
- Optimized dependency management
- Improved handling of modified and deleted patterns during testing


crates/cli/src/commands/patterns_test.rs
```diff
@@ -241,17 +253,26 @@ pub async fn get_marzano_pattern_test_results(
...
         }
     }
-
     Ok(())
 }
 
 pub(crate) async fn run_patterns_test(
     arg: PatternsTestArgs,
     flags: GlobalFormatFlags,
+    watch_mode_args: Option<Vec<String>>,
```
morgante, line 262: This doesn't feel like the right interface.

I think you should probably extract the core / repeated logic into a `test_patterns` function. Then the filtering of which patterns to run is kind of the same whether it's based on watch or exclude or something else.

crates/cli/src/commands/patterns_test.rs
```diff
@@ -276,7 +297,232 @@ pub(crate) async fn run_patterns_test(
...
+    get_marzano_pattern_test_results(testable_patterns, &libs, &arg, flags.clone().into()).await?;
+
+    if arg.watch {
+        enable_watch_mode(arg, flags);
+    }
+    Ok(())
+}
+
+fn enable_watch_mode(cmd_arg: PatternsTestArgs, cmd_flags: GlobalFormatFlags) -> () {
+    let path = Path::new(".grit");
```
morgante, line 309: We don't want to watch all of `.grit` (including .gritmodules). Also this should probably not be created new, but instead should be passed in based on the original patterns retrieved.

crates/cli/src/commands/patterns_test.rs
```diff
@@ -276,7 +297,232 @@ pub(crate) async fn run_patterns_test(
...
+            }
+        }
+    }
+}
+
+async fn get_affected_patterns(modified_file_path: PathBuf) -> Result<Vec<String>> {
+    let modified_file_path_string = modified_file_path.into_os_string().into_string().unwrap();
+
+    let (patterns, _) = resolve_from_cwd(&Source::Local).await?;
+    let affected_patterns = collect_testable_patterns(patterns.clone())
```
morgante, line 360: Rebuilding this from scratch every time is inefficient. We should instead just be looking at the *existing* retrieve patterns that were first found in the original top level command, then filtering those down to the ones related to the modified file.

crates/cli/src/commands/patterns_test.rs
```diff
@@ -276,7 +297,232 @@ pub(crate) async fn run_patterns_test(
...
+    }
+
+    for depen in affected_patterns_dependents_names {
+        if !affected_patterns_names.contains(&depen) {
+            affected_patterns_names.push(depen);
+        }
+    }
+    Ok(affected_patterns_names)
+}
+
```
morgante, line 412: Reparsing shouldn't be necessary. If we need additional metadata as the patterns are built in the first place, collect that in the compiler.

morgante (changes requested):
Thanks for the start, but this should be doable in a cleaner way with less code.

crates/cli/src/commands/patterns_test.rs
Ashu999: I was thinking of using `arg.filter` here, but it only takes single string as input and uses regex. 
If it's ok with you, I can modify the behavior of filter to support multiple arguments, not use regex and filter based on exact pattern name instead. This is how I have seen most of the testing tools work, when targeting a specific test(s).

crates/cli/src/commands/patterns_test.rs
Ashu999: notify crate currently does not support ignoring the directories inside tracked directory(https://github.com/notify-rs/notify/issues/598). I have added a temporary fix to avoid re-running the tests in case any of these files are modified.

crates/cli/src/commands/patterns_test.rs
Ashu999: can you explain this a little more, how to collect the required data in in the compiler? does it happen inside `filter_libs()` in `compiler.rs`?

crates/cli/src/commands/patterns_test.rs
Ashu999: > should be passed in based on the original patterns retrieved
let's say a new pattern is added to a file, we will not know that unless we re-read the patterns from file scratch. right?

crates/cli/src/commands/patterns_test.rs
Ashu999: addressed in: https://github.com/getgrit/gritql/pull/298#discussion_r1614373584

crates/cli/src/commands/patterns_test.rs
morgante: > let's say a new pattern is added to a file, we will not know that unless we re-read the patterns from file. right?

That's true, I wasn't thinking of the case of an entirely new file. This is ok then.

crates/cli/src/commands/patterns_test.rs
morgante: No it doesn't need to be an inner function. You might not need a new function actually, but instead a structure like this:

- `run_patterns_test()`, root that handles computing available patterns, invokes the initial test, and (if configured) calls `enable_watch_mode` *with the patterns found earlier*.
- `enable_watch_mode` listens to notifications, and if files are updated, updates the patterns associated with those files. Importantly, it should *not* invoke `run_patterns_test()` or re-discover patterns - just (a) mutate the patterns based on the file change and (b) invoke `get_marzano_pattern_test_results` itself.
- If necessary, you could have a `do_patterns_test` function that abstracts some of the shared logic from `enable_watch_mode` *and* `run_patterns_test` but it actually looks unnecessary now that I look deeper.

crates/cli/src/commands/patterns_test.rs
morgante: The point is you shouldn't need to reparse any/all patterns to get the dependency tree. Every pattern is already parsed at least one when running the initial test, so you should build the mapping when that happens instead of running again.

crates/gritmodule/src/testing.rs
```diff
@@ -32,6 +33,29 @@ pub fn collect_testable_patterns(
...
+        },
+        local_name: Some(pattern.local_name.clone()),
+    }
+}
+
+pub fn get_grit_pattern_test_info(
+    mut patterns: Vec<ModuleGritPattern>,
+) -> Vec<GritPatternTestInfo> {
+    let testable_patterns: Vec<GritPatternTestInfo> = patterns
+        .iter_mut()
```
morgante, line 52: Use into_iter and avoid needing to clone.

Ashu999: done

crates/cli/src/commands/patterns_test.rs
```diff
@@ -273,11 +287,224 @@ pub(crate) async fn run_patterns_test(
...
+    let path = Path::new(modified_path);
+    let file_patterns = collect_from_file(path, &None).await?;
+    let modified_patterns = get_grit_pattern_test_info(file_patterns);
+
+    let mut modified_pattern_names = <Vec<&String>>::new();
+    for pattern in &modified_patterns {
+        modified_pattern_names.push(pattern.local_name.as_ref().unwrap());
+    }
+    //modified_patterns = patterns which are updated/edited or newly created.
+    //deleted_patterns = patterns which are deleted. Only remaining dependents of deleted_patterns should gets tested.
```
morgante, line 497: Please avoid committing commented out code. /grit new

morgante (changes requested):

crates/cli/src/commands/patterns_test.rs
Ashu999: but these are comments

morgante: Apologies, I was skimming too quickly.

crates/core/src/pattern_compiler/compiler.rs
```diff
@@ -567,6 +567,121 @@ fn find_definition_if_exists(
...
+            } else if n.node.kind() == predicate_call {
+                if let Some(tree) = find_child_tree_definition(
+                    &predicate_file,
+                    parser,
+                    libs,
+                    &mut traversed_stack,
+                    &name,
+                )? {
+                    stack.push(tree);
+                }
```
morgante, line 659: Needing to repeat like this is a code smell. Instead of having a separate lookup for each kind we should be able to collapse into a main map before hand then just traverse that.

crates/core/src/pattern_compiler/compiler.rs
```diff
@@ -567,6 +567,121 @@ fn find_definition_if_exists(
...
+        patterns: pattern_file,
+        predicates: predicate_file,
+        functions: function_file,
+        foreign_functions: foreign_file,
+    } = defs_to_filenames(libs, parser, tree.root_node())?;
+
+    let mut traversed_stack = <Vec<String>>::new();
+
+    // gross but necessary due to running these patterns before and after each file
+    let mut stack: Vec<Tree> = if will_autowrap {
```
morgante, line 593: Just leave this logic out - we don't need to worry about autowrap / before patterns for `--watch`.

.grit/test/test.yaml
```diff
@@ -0,0 +1,82 @@
+# this file is used for integration testing of `grit patterns test --watch`
```
morgante, line 1: Please don't invent a new way to do integration tests. We already have fixtures and existing conventions. Look at them and follow that.

morgante (changes requested):

crates/cli/src/commands/patterns_test.rs
```diff
@@ -273,11 +287,225 @@ pub(crate) async fn run_patterns_test(
 
     let testable_patterns = collect_testable_patterns(patterns);
 
-    if testable_patterns.is_empty() {
-        bail!("No testable patterns found. To test a pattern, make sure it is defined in .grit/grit.yaml or a .md file in your .grit/patterns directory.");
+    get_marzano_pattern_test_results(testable_patterns.clone(), &libs, &arg, flags.clone().into())
```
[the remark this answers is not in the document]
morgante: @Ashu999 greptile has a point here. We should only need to clone when in `arg.watch` mode.

crates/cli/src/commands/patterns_test.rs
```diff
@@ -273,11 +284,193 @@ pub(crate) async fn run_patterns_test(
...
+    // setup debouncer
+    let (tx, rx) = std::sync::mpsc::channel();
+    // notify backend configuration
+    let backend_config = notify::Config::default().with_poll_interval(Duration::from_millis(10));
+    // debouncer configuration
+    let debouncer_config = Config::default()
+        .with_timeout(Duration::from_millis(10))
+        .with_notify_config(backend_config);
+    // select backend via fish operator, here PollWatcher backend
+    let mut debouncer = new_debouncer_opt::<_, notify::PollWatcher>(debouncer_config, tx).unwrap();
```
morgante, line 313: Use `?` where possible.

crates/cli/src/commands/patterns_test.rs
```diff
@@ -273,11 +284,193 @@ pub(crate) async fn run_patterns_test(
...
+    // event processing
+    for result in rx {
+        match result {
+            Ok(event) => {
+                let modified_file_path = event.first().unwrap().path.clone();
+
+                if !modified_file_path.is_file() {
+                    continue;
+                }
+                let modified_file_path = modified_file_path.into_os_string().into_string().unwrap();
```
morgante, line 332: Likely cloning twice unnecessarily.

morgante (changes requested):
Try to be more disciplined with clones.

crates/cli/src/commands/patterns_test.rs
Ashu999: this seems to be the best available way to do it: https://stackoverflow.com/a/42579588

morgante: Right, but the `clone()` on line 327 shouldn't be needed.

Ashu999: yes, it can be changed to:
```rust
                let modified_file_path = &event.first().unwrap().path;

                if !modified_file_path.is_file() {
                    continue;
                }
                let modified_file_path = modified_file_path
                    .clone()
                    .into_os_string()
                    .into_string()
                    .unwrap();
```
which is slightly better because of `if` condition in the middle. but `clone()` can't be avoided as `into_os_string()` requires owned `PathBuf`.

crates/cli_bin/tests/patterns_test.rs
```diff
@@ -309,3 +319,106 @@ fn tests_python_pattern_with_file_name() -> Result<()> {
...
+#[test]
+fn patterns_test_watch_mode_case_patterns_changed() -> Result<()> {
+    let (tx, rx) = mpsc::channel();
+
+    let (temp_dir, temp_grit_dir) = get_fixture(".grit", false)?;
+    let test_yaml_path = temp_grit_dir.join("grit.yaml");
+    let temp_dir_path = temp_dir.path().to_owned();
+
+    let _cmd_handle = thread::spawn(move || {
+        let mut cmd = Command::cargo_bin(BIN_NAME)
```
morgante, line 332: Use `get_test_cmd()` instead and avoid hard-coding the BIN_name.

Ashu999: I did try to use `get_test_cmd()`, but was not able to use `.stdout()` with it. 

was getting below error:
`no method named `stdout` found for mutable reference `&mut assert_cmd::Command` in the current scope
method not found in `&mut Command``

Ashu999: created get_test_process_cmd() to use process::Command, which supports .stdout()

https://github.com/getgrit/gritql/pull/298/commits/58957570dba4313b4980c1d6e6053f3ef861c6dd

morgante: We have many other tests that inspect stdout: https://github.com/getgrit/gritql/blob/f850d3aaaefeefedad2446a230728636632b1484/crates/cli_bin/tests/help.rs#L23

I don't understand why you need a new interface.

morgante (approved):
Once you resolve merge conflicts and tests pass, this is good to go.

morgante:
@Ashu999 Tests and linters are failing. Please make sure everything passes.

PR #309, opened 2024-05-05
fix(js): treat empty statements properly

### LINKED ISSUE #307: Rewrite wrongly changes JavaScript `for` loop conditions
#### Pattern

```grit
engine marzano(0.1)
language js

`$body` => `$body`
```

#### Code before

```javascript
let i = 0;
for (; i < 10; ++i) {
  console.log(i);
}
```

#### Code After

```javascript
let i = 0;
for ( i < 10; ++i) {
  console.log(i);
}
```

#### Expected Behavior

The code shouldn't be modified. The first semicolon in the `for` loop should be preserved.

#### Reproduction

https://app.grit.io/studio?key=BOMSD-cdiDKk_eD23sSAg

### LINKED ISSUE #308: Rewrite wrongly removes the empty statement after the `if` condition in JavaScript.
This issue is similar to #307.

#### Pattern

```grit
engine marzano(0.1)
language js

`$body` => `$body`
```

#### Code before

```javascript
let a = true;
if (a);
console.log(a);
```

#### Code After

```javascript
let a = true;
if (a)
console.log(a);
```

#### Expected Behavior

The code shouldn't be modified. The empty statement after the `if` condition shouldn't be removed. I think this issue also applys to other cases like `else`, `while`, `do`, .etc.

As is suggested by @morgante in https://github.com/getgrit/gritql/issues/307#issuecomment-2094723187, this is also likely caused by the empty statement cleaning up in https://github.com/getgrit/gritql/blob/662b6acd48acbe2190377458bc4dd1933543dc8d/crates/language/src/js_like.rs#L157.

#### Reproduction

https://app.grit.io/studio?key=mBKTvyQw9obvK0czFYbBq

Sec-ant:
Empty statements in JavaScript are not always redundant and removing them regardless will break the code. For example, empty statements (`;`) in the code below are all significant:

```js
for (;;);

if (true);

while (true);
```

This PR treats `empty_statement`s directly under `program` and `statement_block` as redundant ones and they can be removed safely, while the rest are considered as significant ones so we keep them during cleaning up.

Two test cases are added to `crates/core/src/test.rs`.

This PR fixes #307 and also fixes #308.

## Summary by CodeRabbit

- **New Features**
	- Enhanced JavaScript code processing to better handle empty statements, ensuring more accurate code modifications.

- **Tests**
	- Added new tests to verify the correct preservation and removal of empty statements in JavaScript code.

### DIFF
--- a/crates/core/src/test.rs
+++ b/crates/core/src/test.rs
@@ -3884,6 +3884,38 @@ fn js_paren_params() {
 }
 
 #[test]
+fn js_preserve_significant_empty_statements() {
+    run_test_expected({
+        TestArgExpected {
+            pattern: r#"`$body` => `$body`"#.to_owned(),
+            source: r#"
+            for(;;);
+            if(true);
+            while(true);"#
+                .to_owned(),
+            expected: r#"
+            for(;;);
+            if(true);
+            while(true);"#
+                .to_owned(),
+        }
+    })
+    .unwrap();
+}
+
+#[test]
+fn js_remove_redundant_empty_statements() {
+    run_test_expected({
+        TestArgExpected {
+            pattern: r#"`$body` => `$body`"#.to_owned(),
+            source: r#";;{;;};;"#.to_owned(),
+            expected: r#"{}"#.to_owned(),
+        }
+    })
+    .unwrap();
+}
+
+#[test]
 fn array_destrcutring_snippet() {
     run_test_match({
         TestArg {
--- a/crates/language/src/js_like.rs
+++ b/crates/language/src/js_like.rs
@@ -170,6 +170,8 @@ pub(crate) fn jslike_check_replacements(
         && n.text()
             .is_ok_and(|t| ["var", "let", "const"].contains(&t.as_ref()))
         || n.node.kind() == "empty_statement"
+            && n.parent()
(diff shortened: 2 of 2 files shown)


morgante:
@Sec-ant Thanks for the contribution, unfortunately tests are currently failing. You will likely have to handle a few more scenarios where the empty statement is unexpected to get them to pass.

Sec-ant:
The failling test is `output_jsonl`, which I think has nothing to do with this PR because that test case also fails on the main branch.

`output_jsonl` is a snapshot test. Should I update the snapshot in this PR, or create a dedicated PR to fix the test?

morgante:
> The failling test is `output_jsonl`, which I think has nothing to do with this PR because that test case also fails on the main branch: https://github.com/getgrit/gritql/actions/runs/8962305604/job/24611002144#step:10:1
> 
> `output_jsonl` is a snapshot test. Should I update the snapshot in this PR, or create a dedicated PR to fix the test?

I missed that before, it's fine to just update the snapshot in this PR.

Sec-ant:
> I missed that before, it's fine to just update the snapshot in this PR.

Sorry I missed your reply, I created #310 to update it.

Sec-ant:
All the tests are good now.

morgante (approved):
Looks like a good, reasonable solution. Thanks!

PR #321, opened 2024-05-09
chore: update dist config

Gankra:
* set dist=true to override publish=false
* set a 'grit' bin-alias
* add temp names for npm/homebrew packages
* normalize repository urls to match what rust tools expect

I wish there was a good way to do the subdir thing for repository. We could potentially override homepage here as an alternative.


Gankra:
> add temp names for npm/homebrew packages

it's desirable for these to be temp for us to do some test runs / prevent accidents

crates/cli_bin/Cargo.toml
```diff
@@ -56,5 +57,16 @@ grit_tracing = [
 docgen = ["marzano-cli/docgen"]
 
 [package.metadata.dist]
+# this should be published by cargo-dist (needed because publish = false)
+dist = true
+# build flags for published builds
 default-features = false
 features = ["marzano-cli/grit_beta"]
+# the npm package should have this name
+npm-package = "launcher2"
```
morgante, line 66: Would we want this to be `@getgrit/launcher` to replace our existing npm packge?

Gankra: Yeah, I just wasn't sure if you'd want to do a few test runs on a separate package name first -- but we can just set the final one now.

Gankra: i've updated it to "launcher" now

morgante (approved):

PR #333, opened 2024-05-14
feat: use axoupdater for cargo-dist built binaries

Gankra:
## Summary by CodeRabbit

- **New Features**
  - Introduced the Grit CLI, a developer tool for simplifying software maintenance.
  - Added functionality for checking and installing the latest releases using GitHub.

- **Updates**
  - Updated npm package name from "launcher" to "cli".
  - Updated Homebrew package name from "grit" to "grit_beta".

- **Documentation**
  - Added README with details on Grit CLI, its products, documentation, and distribution methods.

### DIFF
--- a/crates/cli/src/updater.rs
+++ b/crates/cli/src/updater.rs
@@ -1,5 +1,6 @@
 use anyhow::bail;
 use anyhow::{Context, Result};
+use axoupdater::{AxoUpdater, ReleaseSource, ReleaseSourceType, Version};
 use chrono::{DateTime, NaiveDateTime, Utc};
 use colored::Colorize;
 use futures_util::StreamExt;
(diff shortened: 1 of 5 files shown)


crates/cli/src/updater.rs
```diff
@@ -300,18 +301,16 @@ impl Updater {
...
+                    let (result, version) = match app {
+                        SupportedApp::Marzano | SupportedApp::Gouda => {
+                            check_release_axo(app, &current_binary).await.unwrap()
+                        }
+                        _ => check_release_internal(app, &current_binary).await.unwrap(),
+                    };
+                    if result {
                         let mut found = found_updates.lock().await;
-                        found.push((app, manifest.version.unwrap()));
+                        found.push((app, version.unwrap()));
```
morgante, line 313: Since this method returns a Result<> we should avoid unwrap - can you refactor to remove?

morgante (changes requested):

crates/cli/src/updater.rs
mistydemeo: Sure thing! I've pushed a commit which replicates the original logic, with a `return` if an `Err` is encountered when checking for a release. This is within an `async` block, which wasn't itself returning `Result<>`, so the question mark operator was trickier.

The `version.unwrap()`, which is on an `Option<>` is left alone from how it was before this PR, but let me know if you'd like me to adjust that as well.

mistydemeo: Now that I think about it, eliminating the `version.unwrap()` doesn't add too much more complexity, so pushed a change that does that as well.

morgante: Great, thanks for improving it.

morgante (approved):

PR #366, opened 2024-06-01
refactor: remove `AstNode::full_source()`

arendjr:
This PR removes the `full_source()` from the `AstNode` trait and the `NodeWithSource` implementation. The motivation for this change is that the method proved problematic to implement with Biome's CST. Conceptually the method was of course also a bit strange, since it asks an individual node to retrieve the source of the entire tree.

Fortunately, the method was only used in 2 places:
- Inside `apply_effects()`. This one was relatively easy to update, since the function was only called with the root node anyway. So instead of passing a node, we now pass the tree. A new `source()` method was added to the `Ast` trait, where it makes more sense.
- For grabbing the source to pass to `adjust_padding()`. This one was more tricky to resolve, and the bulk of this PR is in the changes it caused. To resolve it, I removed `adjust_padding()` as it existed and added a new `align_padding()` to the `Language` trait (the semantics remained the same, but the rename seemed more fitting). Apart from moving the method, its signature changed to receive a `Node` instead of `src: &str`. For the Marzano implementation this seems a trivial change, since it simply grabs the source from the `Node` anyway, but it will allow the Biome implementation to use CST traversal to determine the surrounding padding.

Together with `adjust_padding()`, `pad_snippet()` was moved as well. And because of the added methods on the `Language` trait, all the language implementations needed updates as well. I've added a few small macros to streamline their implementations.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new method `contains` in the `CodeRange` struct for checking if an index is within a range.
  - Added `EffectKind` enum with variants `Rewrite` and `Insert`.

- **Refactor**
  - Updated version to "0.3.0" and added "Grit Contributors" to authors list.
  - Reorganized imports and method calls for padding snippets.
  - Refactored `MarzanoContext` and `ExecContext` method signatures.
  - Replaced local definitions of `EffectKind` with imports from `grit_util`.
  - Introduced macros for delegate usage in language implementations.

- **Bug Fixes**
  - Removed outdated methods and functions related to padding and effect ranges, ensuring better alignment and processing.

- **Chores**
  - Updated dependencies across multiple crates to ensure compatibility with new versions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1400,7 +1400,7 @@ dependencies = [
 
 [[package]]
 name = "grit-pattern-matcher"
-version = "0.2.0"
+version = "0.3.0"
 dependencies = [
  "anyhow",
  "elsa",
@@ -1414,7 +1414,7 @@ dependencies = [
 
 [[package]]
 name = "grit-util"
-version = "0.2.0"
+version = "0.3.0"
 dependencies = [
  "derive_builder",
  "napi",
@@ -1426,7 +1426,7 @@ dependencies = [
 
 [[package]]
 name = "grit-wasm-bindings"
-version = "0.2.0"
+version = "0.3.0"
 dependencies = [
  "ai_builtins",
  "anyhow",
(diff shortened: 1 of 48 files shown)


arendjr:
I've also bumped the version for the `grit-util` and `grit-pattern-matcher` crates, so they can be released again.

morgante (approved):

PR #418, opened 2024-07-13
fix: removal of trailing comma lines (#416)

### LINKED ISSUE #416: Hanging commas after removing kwargs in Python
Continuing from Discord [message](https://discord.com/channels/1140989670751219762/1140989757392953424/1261006232618143877).

```grit
`TaskMetadata($args)` where {
    $args <: any {
        contains `n_samples=$_` as $ns_kwarg where {
            $ns_kwarg <: `n_samples = $ns_val` => .
        },
        contains keyword_argument(name="avg_character_length") as $avg_kwarg where {
            $avg_kwarg <: `avg_character_length = $avg_val` => `stats=GeneralDescriptiveStats(n_samples=$ns_val, avg_character_length=$avg_val)`
        },
    },
}
```

```python
from pydantic import BaseModel


class TaskMetadata(BaseModel):
    n_samples: dict[str, int]
    avg_character_length: dict[str, float]


if __name__ == "__main__":
    TaskMetadata(
        name="TbilisiCityHallBitextMining",
        dataset={
            "path": "jupyterjazz/tbilisi-city-hall-titles",
            "revision": "798bb599140565cca2dab8473035fa167e5ee602",
        },
        description="Parallel news titles from the Tbilisi City Hall website (https://tbilisi.gov.ge/).",
        type="BitextMining",
        category="s2s",
        eval_splits=[_EVAL_SPLIT],
        eval_langs=_EVAL_LANGS,
        main_score="f1",
        domains=["News"],
        text_creation="created",
        n_samples={_EVAL_SPLIT: 1820},
        reference="https://huggingface.co/datasets/jupyterjazz/tbilisi-city-hall-titles",
        date=("2024-05-02", "2024-05-03"),
        form=["written"],
        task_subtypes=[],
        license="Not specified",
        socioeconomic_status="mixed",
        annotations_creators="derived",
        dialect=[],
        bibtex_citation="",
        avg_character_length={_EVAL_SPLIT: 78},
    )
```

Morgante suggested it might be a bug somewhere around [here](https://github.com/getgrit/gritql/blob/9587e9e5558c583d69514c1f581f8244ac551616/crates/core/src/inline_snippets.rs#L298).

Results in a dangling comma:
image copy

abhishek818:
/claim #416 
/closes #416

### DIFF
--- a/crates/core/src/inline_snippets.rs
+++ b/crates/core/src/inline_snippets.rs
@@ -234,12 +234,28 @@ pub(crate) fn inline_sorted_snippets_with_offset(
         }
     }
 
+    let mut trailing_comma_lines = vec![];
+
     for (range, snippet) in replacements {
         let range = adjust_range(&range.effective_range(), offset, &code)?;
         if range.start > code.len() || range.end > code.len() {
             bail!("Range {:?} is out of bounds for code:\n{}\n", range, code);
         }
-        code.replace_range(range, snippet);
+        code.replace_range(range.clone(), snippet);
+
+        // Check if the replacement results in any lines containing just a single comma
+        let line_start = code[..range.start].rfind('\n').map_or(0, |pos| pos + 1);
+        let line_end = code[range.start..].find('\n').map_or(code.len(), |pos| range.start + pos);
+
+        let line_content = code[line_start..line_end].trim();
+        if line_content == "," {
+            trailing_comma_lines.push(line_start..line_end+1);
+        }
+    }
+
+    // Remove lines with a trailing comma
+    for range in trailing_comma_lines {
+        code.replace_range(range, "");
     }
 
     Ok((code, output_ranges, original_ranges))
--- a/crates/core/src/test.rs
+++ b/crates/core/src/test.rs
@@ -12014,6 +12014,104 @@ fn trailing_comma_import_from_python_with_alias() {
     .unwrap();
 }
 
+// refer https://github.com/getgrit/gritql/issues/416
+#[test]
+fn trailing_comma_after_argument_removal() {
+    run_test_expected({
+        TestArgExpected {
+            pattern: r#"
+                language python
+                `TaskMetadata($args)` where {
+                        $args <: any {
+                            contains `n_samples=$_` as $ns_kwarg where {
+                                $ns_kwarg <: `n_samples = $ns_val` => .
+                            },
+                            contains `avg_character_length=$_` as $avg_kwarg where {
+                                $avg_kwarg <: `avg_character_length = $avg_val` => `stats=GeneralDescriptiveStats(n_samples=$ns_val, avg_character_length=$avg_val)`
+                            },
+                        },
+                    }
+                "#.to_owned(),
+            source: r#"
+                from pydantic import BaseModel
+
+
+                class TaskMetadata(BaseModel):
+                    n_samples: dict[str, int]
+                    avg_character_length: dict[str, float]
+
+
+                if __name__ == "__main__":
+                    TaskMetadata(
+                        name="TbilisiCityHallBitextMining",
+                        dataset={
+                            "path": "jupyterjazz/tbilisi-city-hall-titles",
+                            "revision": "798bb599140565cca2dab8473035fa167e5ee602",
+                        },
+                        description="Parallel news titles from the Tbilisi City Hall website (https://tbilisi.gov.ge/).",
+                        type="BitextMining",
+                        category="s2s",
+                        eval_splits=[_EVAL_SPLIT],
+                        eval_langs=_EVAL_LANGS,
+                        main_score="f1",
+                        domains=["News"],
+                        text_creation="created",
+                        n_samples={_EVAL_SPLIT: 1820},
+                        reference="https://huggingface.co/datasets/jupyterjazz/tbilisi-city-hall-titles",
+                        date=("2024-05-02", "2024-05-03"),
+                        form=["written"],
+                        task_subtypes=[],
+                        license="Not specified",
+                        socioeconomic_status="mixed",
+                        annotations_creators="derived",
+                        dialect=[],
+                        bibtex_citation="",
+                        avg_character_length={_EVAL_SPLIT: 78},
+                    )
+            "#
+                .to_owned(),
+            expected: r#"
+                from pydantic import BaseModel
+
+
+                class TaskMetadata(BaseModel):
+                    n_samples: dict[str, int]
+                    avg_character_length: dict[str, float]
+
+
+                if __name__ == "__main__":
+                    TaskMetadata(
+                        name="TbilisiCityHallBitextMining",
+                        dataset={
+                            "path": "jupyterjazz/tbilisi-city-hall-titles",
+                            "revision": "798bb599140565cca2dab8473035fa167e5ee602",
+                        },
+                        description="Parallel news titles from the Tbilisi City Hall website (https://tbilisi.gov.ge/).",
+                        type="BitextMining",
+                        category="s2s",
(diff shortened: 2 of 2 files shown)


abhishek818:
@morgante Pls review. All tests are passing locally for me.

crates/core/src/inline_snippets.rs
```diff
@@ -234,12 +234,28 @@ pub(crate) fn inline_sorted_snippets_with_offset(
         }
     }
 
+    let mut trailing_comma_lines = vec![];
```
morgante, line 237: We already have a `delete_hanging_comma` function that should be used - we don't want two different pieces of code dealing with orphaned commas.

Any solution should either be [there](https://github.com/getgrit/gritql/blob/9587e9e5558c583d69514c1f581f8244ac551616/crates/core/src/inline_snippets.rs#L298-L374) or in a [replacement](https://github.com/getgrit/gritql/blob/9587e9e5558c583d69514c1f581f8244ac551616/crates/language/src/python.rs#L80).

morgante (changes requested):

crates/core/src/inline_snippets.rs
abhishek818: @morgante 
Had tried that earlier. I mean, placing the code fix in **delete_hanging_comma** function or in sub functions called inside it. That resulted in failure of an existing edge test case. This happens since **offsets** are again calculated after calling delete_hanging_comma and then actual replacements happens.

So, I feel doing this fix during actual replacements is the best approach.

Pasting here the change I was trying with :
just manipulated the range indexes here.
  
```
fn delete_hanging_comma(
    code: &str,
    replacements: &mut [(EffectRange, String)],
    offset: usize,
) -> Result<(String, Vec<ReplacementInfo>)> {
    // Handle the case when after applying replacements, a single comma
    // gets left behind on a single line
    let mut temp_code = code.to_string();
    for (range, snippet) in replacements.iter_mut() {
        let mut adjusted_range = adjust_range(&range.effective_range(), offset, &temp_code)?;
        if adjusted_range.start > temp_code.len() || adjusted_range.end > temp_code.len() {
            bail!("Range {:?} is out of bounds for code:\n{}\n", adjusted_range, temp_code);
        }

        temp_code.replace_range(adjusted_range.clone(), snippet);

        let line_start = temp_code[..adjusted_range.start].rfind('\n').map_or(0, |pos| pos + 1);
        let line_end = temp_code[adjusted_range.start..].find('\n').map_or(temp_code.len(), |pos| adjusted_range.start + pos);
        let line_content = temp_code[line_start..line_end].trim();
        
        // Check for lines with a single comma after replacements
        if line_content == "," {
            // Ensure the next character is a newline
            if line_end < temp_code.len() && temp_code.chars().nth(line_end) == Some('\n') {
                adjusted_range.end = line_end + 1;
            }

            // Update the original range to replace the range with an empty string later.
            range.range.end += adjusted_range.end - line_start;
        }
    }
```

morgante: You should be able to do it in one of those two locations. Please don't make our handling even more spaghetti.

abhishek818: @morgante  hey one doubt I have :

In below test case (existing), shouldnt we remove the empty newline?

```
#[test]
fn only_deduplicate_rewrites6() {
    run_test_expected({
        TestArgExpected {
            pattern: r#"
                engine marzano(0.1)
                language json
                pair($key, $value) => . where {
                    $key <: `"extends"`,
                }"#
            .to_owned(),
            source: r#"
            {
                "env": "bar",
                "extends": "foo",
                "baz": 1
            }"#
            .to_owned(),
            expected: r#"
            {
                "env": "bar",
                                                     <----------------------------------
                "baz": 1
            }"#
            .to_owned(),
        }
    })
    .unwrap();
}
```

morgante: No, it is not necessary for Grit to handle removing newlines. We rely on formatters to handle that.

abhishek818: @morgante got it. This fix should be satisfactory, have a look..

crates/core/src/inline_snippets.rs
```diff
@@ -300,6 +300,27 @@ fn delete_hanging_comma(
     replacements: &mut [(EffectRange, String)],
     offset: usize,
 ) -> Result<(String, Vec<ReplacementInfo>)> {
+    // Handle the case when after applying replacements, a single comma
```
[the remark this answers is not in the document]
abhishek818: @morgante hey, i tried with your given pattern on [https://app.grit.io/studio?key=-Pm14wW6mvH0S41J_hFh8](https://app.grit.io/studio?key=-Pm14wW6mvH0S41J_hFh8), it worked. 

Since I am still new to gritql and its rules, can you give a example where it fails. Do you mean presence of matching keywords on multiple lines ? or presence of keyword on the first line (that works) ..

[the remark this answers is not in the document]
abhishek818: @morgante 
you mean when it is like this:
```
bibtex_citation="",
n_samples={_EVAL_SPLIT: 1820}, avg_character_length={_EVAL_SPLIT: 78},
```

then my current fix prints with a extra space:
```
bibtex_citation="",
 stats=GeneralDescriptiveStats(n_samples={_EVAL_SPLIT: 1820}, avg_character_length={_EVAL_SPLIT: 78}),
```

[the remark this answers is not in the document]
abhishek818: I don't find any other edge cases,
Since I am still new to gritql, can you provide a case where it fails.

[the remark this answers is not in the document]
morgante: Your fix completely fails in cases where the arguments are on 1 line. *The original bug isn't related to line handling at all and introducing that is a fundamental mistake in your fix.*

```

/// Same as above test, but ensures the behavior doesn't depend on line breaks
#[test]
fn trailing_comma_after_argument_removal_one_line() {
    run_test_expected({
        TestArgExpected {
            pattern: r#"
                language python
                `TaskMetadata($args)` where {
                        $args <: any {
                            contains `n_samples=$_` as $ns_kwarg where {
                                $ns_kwarg <: `n_samples = $ns_val` => .
                            },
                            contains `avg_character_length=$_` as $avg_kwarg where {
                                $avg_kwarg <: `avg_character_length = $avg_val` => `stats=GeneralDescriptiveStats(n_samples=$ns_val, avg_character_length=$avg_val)`
                            },
                        },
                    }
                "#.to_owned(),
            source: r#"
                TaskMetadata(description="Parallel news titles from the Tbilisi City Hall website (https://tbilisi.gov.ge/).", main_score="f1", domains=["News"], text_creation="created", n_samples={_EVAL_SPLIT: 1820}, reference="https://huggingface.co/datasets/jupyterjazz/tbilisi-city-hall-titles")
            "#
                .to_owned(),
            expected: r#"
                TaskMetadata(description="Parallel news titles from the Tbilisi City Hall website (https://tbilisi.gov.ge/).", main_score="f1", domains=["News"], text_creation="created",   reference="https://huggingface.co/datasets/jupyterjazz/tbilisi-city-hall-titles")
            "#
                .to_owned(),
        }
    })
    .unwrap();
}
```

Also has nothing to do with GritQL, you don't need to change the pattern to see how this is broken.

morgante (changes requested):
Don't rely on line breaks.

morgante:
I think you're probably working in entirely the wrong place.

The bug is likely in the mini state machine / parser we have here: https://github.com/getgrit/rewriter/blob/main/vendor/gritql/crates/core/src/inline_snippets.rs#L535-L652

PR #419, opened 2024-07-16
fix: try using replacement approach

morgante:
https://github.com/getgrit/gritql/issues/416, but unfortunately the comma doesn't appear as an error node.


morgante:
Done in https://github.com/getgrit/gritql/pull/566

PR #453, opened 2024-08-09
refactor: use `thiserror` instead of `anyhow`

### LINKED ISSUE #409: Request to change `anyhow` to `thiserror` (or custom error) in `grit-pattern-matcher`
Mostly because `anyhow` is meant for [application code](https://github.com/dtolnay/anyhow?tab=readme-ov-file#comparison-to-thiserror), and this forces downstream dependencies (Biome for example) to use `anyhow` to import the types and correctly handle the errors.

I wonder if you're open to changing this for a custom error or us the library `thiserror`. I am happy to send a PR too.

morgante:
fixes #409

### DIFF
--- a/crates/cli/src/commands/apply_pattern.rs
+++ b/crates/cli/src/commands/apply_pattern.rs
@@ -269,7 +269,7 @@ pub(crate) async fn run_apply_pattern(
     }
     #[cfg(not(feature = "ai_querygen"))]
     if arg.ai {
-        bail!("Natural language processing is not enabled in this build");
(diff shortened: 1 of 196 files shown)


crates/cli/src/commands/apply_pattern.rs
```diff
@@ -269,7 +269,7 @@ pub(crate) async fn run_apply_pattern(
     }
     #[cfg(not(feature = "ai_querygen"))]
     if arg.ai {
-        bail!("Natural language processing is not enabled in this build");
```
morgante, line 272: This is the kind of change that seems pointless.

PR #462, opened 2024-08-15
Enabling Grit metavariables inside of flow maps

dscleaver:
The snippet `` `{ key: $value }` `` will now match with yaml of the form `{
key: value }`


morgante (approved):
Looks great, and thanks for including tests!

PR #482, opened 2024-09-01
feat : aliases for --lang

### LINKED ISSUE #445: Aliases for `--lang`
I'd like to be able to use aliases like `py` for the lang argument:

```
grit apply --lang py 'identifier()'
error: invalid value 'py' for '--language <LANGUAGE>'
  [possible values: js, html, css, json, java, csharp, python, markdown, go, rust, ruby, solidity, hcl, yaml, sql, vue, toml, php, phponly]

  tip: a similar value exists: 'python'
```

itsdheerajdp:
Fixes #445 
/claim #445

## Summary by CodeRabbit

- **New Features**
	- Enhanced language identifier handling with normalization for improved input flexibility and accuracy.
- **Bug Fixes**
	- Reduced errors from unrecognized language identifiers by introducing a fallback mechanism.

### DIFF
--- a/crates/lsp/src/language.rs
+++ b/crates/lsp/src/language.rs
@@ -31,7 +31,7 @@ pub fn language_id_to_pattern_language(language_id: &str) -> Option<PatternLangu
         }
         "html" => Some(PatternLanguage::Html),
         "css" => Some(PatternLanguage::Css),
-        "json" => Some(PatternLanguage::Json),
+        "json" | "jsonc"  => Some(PatternLanguage::Json),
         "java" => Some(PatternLanguage::Java),
         "csharp" => Some(PatternLanguage::CSharp),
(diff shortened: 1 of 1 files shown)


morgante (commented):
Please add a test case.

crates/lsp/src/language.rs
```diff
@@ -1,13 +1,37 @@
 use marzano_language::target_language::{PatternLanguage, TargetLanguage};
 
 pub fn language_id_to_pattern_language(language_id: &str) -> Option<PatternLanguage> {
```
morgante, line 3: This is used specifically for vs code. It should not be used for CLI stuff... I am 90% sure your change didn't fix the issue / wasn't tested.

morgante:
This isn't remotely correct.

Issue #168, opened 2024-04-06
Switch auto-formatting from prettier to Biome


morgante:
Currently, we [use prettier](https://github.com/getgrit/gritql/blob/f7a8d8670425981d6c87595009338e8b903aa29c/crates/gritmodule/src/formatting.rs#L193) for auto-formatting JS unit test cases.

We should switch to [Biome](https://github.com/biomejs/biome) because:
- It is faster
- It doesn't require any external dependencies, like a node environment
- Biome is partnering with us on the plugin system

PR #487, opened 2024-09-05
WIP: Add support for biome formatter (#168)

xrd:
(re: https://github.com/getgrit/gritql/issues/168)

## Usage:

Use biome formatter instead of prettier. 

- [ ] TODO: Please resolve questions below.

## Altering run time behavior

Extra variables which can be used to alter behavior:

`FORCE_PRETTIER=1` will force grit to use prettier instead.
`FORCE_NPX_BIOME=1` will force grit to use biome via npx.

## Questions:

- [ ] Issue indicates biome is better because it requires no nodejs runtime. 
  - [ ] Does this mean we want to guarantee user has biome already installed and `grit doctor` will check that prerequisite?
- [ ] Verify that we want to use biome for JS/TS/Json and continue to use prettier for CSS/HTML/Markdown/YAML as per https://biomejs.dev/internals/language-support/
- [ ] Do we care about providing an escape hatch if a user wants to continue using prettier (`FORCE_PRETTIER=1`)? If not, these other questions are moot!
- [ ] If `FORCE_PRETTIER=1` is added to the invocation, a user can elect to run `npx prettier...` as the formatter instead of biome. Should we add another envvar which permits them to use the system prettier in case they already have it?
- [ ] I did not see a way to extend the grit.yaml configuration file which seems like a better place to put configuration settings like the envvars. Is the config file the right place to add these configuration settings? How can those be globally specified for all projects if that's what a user wants?
- [ ]  What if a user has team defined settings for prettier or biome that need to be passed in when formatting code? Can we support that? Biome support an additional config file using `BIOME_CONFIG_PATH` so they can run grit and override configuration that way. Does not seem like prettier supports adding config path specification to envvars.
- [ ] Should we add `FORCE_SYSTEM_PRETTIER=1` to allow user to choose system installed prettier? This is inconsistent with the other variables because prettier is using npx in the default case but switching to biome means we use a system installed version instead of npx download.

### DIFF
--- a/crates/gritmodule/src/formatting.rs
+++ b/crates/gritmodule/src/formatting.rs
@@ -2,12 +2,77 @@ use anyhow::{Context, Result};
 use log::{debug, info};
 use marzano_language::target_language::PatternLanguage;
 use marzano_util::rich_path::RichFile;
+use std::env;
 use tempfile::TempDir;
 use tokio::{
     fs::{self, create_dir_all, OpenOptions},
     io::AsyncWriteExt,
     process::Command,
 };
+
+async fn preferred_formatter(dir: &TempDir) {
+    log::debug!("Running code formatter");
+    match env::var("FORCE_PRETTIER") {
+        Ok(s) => {
+            if s == "1" {
+                prettier(dir, true).await;
+            } else {
+                biome(dir).await;
+            }
+        }
+        Err(_) => biome(dir).await,
+    }
+}
+
+async fn biome(dir: &TempDir) {
+    match env::var("FORCE_NPX_BIOME") {
+        Ok(v) => {
+            if v == "1" {
+                run_npx_biome(dir).await;
+            } else {
+                run_system_biome(dir).await;
+            }
+        }
+        Err(_) => run_system_biome(dir).await,
+    }
+}
+
+async fn run_npx_biome(dir: &TempDir) {
+    run_biome(dir, vec!["npx", "@biomejs/biome", "format", "--write"]).await;
+}
+
+async fn run_system_biome(dir: &TempDir) {
+    run_biome(dir, vec!["biome", "format", "--write"]).await
+}
+
+async fn run_biome(dir: &TempDir, commands: Vec<&str>) {
+    let mut cmd = Command::new(&commands[0]);
+    for a in &commands[1..] {
+        cmd.arg(a);
+    }
+    cmd.arg(dir.path().join("**/*"));
+    let output = cmd.output().await;
+    log::debug!("biome output: {:?}", output)
+}
+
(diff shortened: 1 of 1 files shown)


morgante:
>  Does this mean we want to guarantee user has biome already installed and grit doctor will check that prerequisite?

No, like with other formatters we should just skip formatting if `biome` isn't present.

>  Verify that we want to use biome for JS/TS/Json and continue to use prettier for CSS/HTML/Markdown/YAML as per https://biomejs.dev/internals/language-support/

Yes.

>  Do we care about providing an escape hatch if a user wants to continue using prettier (FORCE_PRETTIER=1)? If not, these other questions are moot!

No, we should not have both.

morgante:
Code looks goodish but tests aren't passing - you probably need to update the GitHub actions to install Biome.

xrd:
I added biome to the actions but this appears to require your approval. Do
you want me run these as a PR in my own fork so you avoid using your runner
credits?

morgante:
I don't mind clicking approve, but you should also be able to run tests locally. It looks like some snapshots will need updating, and the standard library tests will also need Biome installed.

xrd:
I got the snapshot tests fixed in the gritql repo. I love those snapshot tests.

As a quick aside, I wasn't sure if there is a recommended way to use my local development version of grit/marzano, so I just added it to my path. If this is dumb for some reason, please let me know, but it appears to use the new biome formatter.

$ cd Projects/grit.io/stdlib
$ export PATH=~/Projects/grit.io/gritql/target/debug:$PATH
$ marzano patterns test

When I ran the stdlib tests, I see a lot of changes. At first glance they appear to be mostly quote issues (single quote to double quote) and missing semicolon issues, for the most part. 

I see two obvious paths: configure biome to match prettier defaults, or fix the tests to match biome preferences. Do you have an opinion on this? There are some other formatting changes in the react/TSX code so this diff might get big regardless, or the biome config might get big, depending on which way we go.

Here is a gist with the stdlib test results:

https://gist.github.com/xrd/aafb4eb68209e553a6c8c79abe12dae6

morgante:
Biome should actually match prettier defaults. From the diff, it looks like you're not actually successfully running the formatter at all. Ex. this code is clearly unformatted:

```
  ✗ Existing import append
     import { posthog } from '../../services/flags';
     import { other, sanitizeFilePath } from '@getgrit/universal';
     import type { MarzanoResolvedPattern } from '@getgrit/sdk';
    -import { marzanoResolvedPatternToResolvedGritPattern } from '@getgrit/sdk';
    +import { marzanoResolvedPatternToResolvedGritPattern  } from '@getgrit/sdk';
     import path from 'path';
     import { ApplicationFailure } from '@temporalio/workflow';
 ```

PR #488, opened 2024-09-06
Aliases for --lang.

xonx4l:
Issue-: #445

Aliases for --lang.

Added the from_string_or_alias function to handle both full language names and their aliases.

### DIFF
--- a/crates/language/src/target_language.rs
+++ b/crates/language/src/target_language.rs
@@ -129,6 +129,48 @@ impl PatternLanguage {
         Self::from_string(lang, flavor.as_deref())
     }
 
+impl PatternLanguage {
+    pub fn from_string_or_alias(name: &str, flavor: Option<&str>) -> Option<Self> {
+        match name {
+            "py" | "python" => Some(Self::Python),
+            "js" | "javascript" => match flavor {
+                Some("jsx") => Some(Self::Tsx),
+                Some("flow") => Some(Self::Tsx),
+                Some("FlowComments") => Some(Self::Tsx),
+                Some("typescript") => Some(Self::TypeScript),
+                Some("js_do_not_use") => Some(Self::JavaScript),
+                _ => Some(Self::Tsx),
+           },
+           "ts" | "typescript" => Some(Self::Typescript),
+           "html" => Some(Self::Html),
+           "css" => Some(Self::Css),
+           "json" => Some(Self::Json),
+           "java" => Some(Self::Java),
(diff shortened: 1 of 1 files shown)


crates/language/src/target_language.rs
```diff
@@ -245,6 +288,7 @@ impl PatternLanguage {
     }
 
     pub fn from_extension(extension: &str) -> Option<Self> {
+        Self::from_string_or_alias(extension, None)
```
morgante, line 291: This call isn't doing anything, I think you probably meant to delegate this entirely?

crates/language/src/target_language.rs
```diff
@@ -146,6 +188,7 @@
     }
 
     pub fn from_string(name: &str, flavor: Option<&str>) -> Option<Self> {
+        Self::from_string_or_alias(name, flavor)
```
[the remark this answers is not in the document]
morgante: I don't think we need both from_string and from_string_or_alias - it is fine to just insert the aliases into here.

morgante (commented):
Thanks! I'm not sure this actually fixes the issue because clap attempts to use the enum directly. Can you add CLI test in `crates/cli_bin/tests/apply.rs` too?

crates/language/src/target_language.rs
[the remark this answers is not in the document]
xonx4l: You're right. We can simplify the implementation by merging the functionality of from_string and from_string_or_alias into a single from_string function.

crates/language/src/target_language.rs
xonx4l: I have changed  the from_extension function which now uses the updated from_string function, which  handles both full language names and aliases.

crates/language/src/target_language.rs
```diff
@@ -129,6 +129,48 @@ impl PatternLanguage {
         Self::from_string(lang, flavor.as_deref())
     }
 
+impl PatternLanguage {
+    pub fn from_string_or_alias(name: &str, flavor: Option<&str>) -> Option<Self> {
```
morgante, line 133: @xonx4l This should be removed, you don't need both.

xonx4l: Done!!!

xonx4l:
> Thanks! I'm not sure this actually fixes the issue because clap attempts to use the enum directly. Can you add CLI test in `crates/cli_bin/tests/apply.rs` too?

working on it!!!

PR #536, opened 2024-10-09
perf: remove `im` crate

arendjr:
Gives rougly a 25-30% performance improvement on Biome's Grit benchmarks.

Heads-up: This changes a signature in [api.rs](https://github.com/getgrit/gritql/compare/main...arendjr:remove-im?expand=1#diff-2290d10881b8010042083f2b283329457b3f5f321cda8a1747e8bf0ae707a1ff). It looks like it should be easy enough for you to adapt, but let me know if you prefer me to revert to `im` there.

There's a bunch of `.iter()` => `as usize` changes that were unfortunately necessary due to ambiguities.

Would be great if 0.4 of the crates can be released after merging this: https://github.com/getgrit/gritql/pull/498

## Greptile Summary

--
API usage limit has reached your account's monthly budget. API requests will be rejected.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -318,15 +318,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ed570934406eb16438a4e976b1b4500774099c13b8cb96eec99f620f05090ddf"
 
 [[package]]
-name = "bitmaps"
-version = "2.1.0"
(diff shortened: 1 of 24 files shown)


morgante:
Thanks, I've long wanted to remove this and it should be easy enough to adapt internally too.

morgante (approved):

PR #544, opened 2024-10-20
fix: 445 `--lang` aliases

### LINKED ISSUE #445: Aliases for `--lang`

Alex-ley-scrub:
@algora-pbc /claim #445 

- test(apply): add simple stdin example from documentation as test case with an existing language and a language alias, which will fail until implemented
  - see #445 
- fix(target_language): allow language aliases and extensions to be used and also map to the correct `PatternLanguage`
  - fixes #445 

verified as fixed with (fails before PR commits succeeds after):
```
echo 'console.log(hello)' | cargo run -- apply '`hello` => `goodbye`' --stdin --lang=javascript
```
two test cases added to cover this in `crates/cli_bin/tests/apply.rs` which also verifies fix (fails before PR commits succeeds after - `cargo test --test apply`)

## Greptile Summary

--

### DIFF
--- a/crates/cli/src/commands/apply_pattern.rs
+++ b/crates/cli/src/commands/apply_pattern.rs
@@ -159,7 +159,7 @@ pub struct ApplyPatternArgs {
     #[clap(long)]
     ai: bool,
     /// Change the default language to use for the pattern (if unset, JavaScript is used by default)
-    #[clap(long = "language", alias = "lang")]
+    #[clap(long = "language", alias = "lang", value_parser = PatternLanguage::value_parser())]
     pub language: Option<PatternLanguage>,
 }
 
--- a/crates/cli/src/commands/apply_pattern.rs
+++ b/crates/cli/src/commands/apply_pattern.rs
@@ -159,7 +159,7 @@ pub struct ApplyPatternArgs {
     #[clap(long)]
     ai: bool,
     /// Change the default language to use for the pattern (if unset, JavaScript is used by default)
-    #[clap(long = "language", alias = "lang", value_parser = PatternLanguage::value_parser())]
+    #[clap(long = "language", alias = "lang")]
     pub language: Option<PatternLanguage>,
 }
 
--- a/crates/language/src/target_language.rs
+++ b/crates/language/src/target_language.rs
@@ -27,6 +27,7 @@ use crate::{
     yaml::Yaml,
 };
 use anyhow::Result;
+use clap::builder::{MapValueParser, PossibleValuesParser, TypedValueParser};
 use clap::ValueEnum;
 use grit_util::Order;
 use grit_util::{Ast, AstNode, ByteRange, CodeRange, Language, Parser, SnippetTree};
@@ -81,6 +82,17 @@ pub enum PatternLanguage {
     #[value(skip)]
     Universal,
 }
+
+// https://github.com/getgrit/gritql/issues/445
+// add any more language aliases here
+// you don't need to add the primary name, nor the extensions
+static LANGUAGE_ALIASES: &[(PatternLanguage, &'static [&'static str])] = &[
+    (PatternLanguage::JavaScript, &["javascript"]),
+    (PatternLanguage::TypeScript, &["typescript"]),
+    (PatternLanguage::Tsx, &["flow"]),
+    (PatternLanguage::Sql, &["mysql", "postgresql"]),
+    (PatternLanguage::PhpOnly, &["phponly"]),
+];
 
 impl fmt::Display for PatternLanguage {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
@@ -171,6 +183,88 @@ impl PatternLanguage {
     pub fn get_language(src: &str) -> Option<Self> {
         let mut parser = MarzanoGritParser::new().unwrap();
         Self::get_language_with_parser(&mut parser, src)
+    }
+
+    fn get_aliases(&self) -> &'static [&'static str] {
+        for (variant, aliases) in LANGUAGE_ALIASES {
+            if variant == self {
+                return aliases;
+            }
+        }
+        &[]
+    }
+
+    fn map_value(s: String) -> Self {
+        if let Some(lang) = Self::from_string(&s, None) {
+            return lang;
+        }
+        if let Some(lang) = Self::from_extension(&s) {
+            return lang;
+        }
+        let s = s.to_lowercase();
+        for (variant, aliases) in LANGUAGE_ALIASES {
+            for alias in *aliases {
+                if *alias == s {
+                    return *variant;
+                }
+            }
+        }
+        Self::Universal
+    }
+
+    fn possible_values() -> Vec<&'static str> {
+        let mut unique_variants = std::collections::HashSet::new();
+        let mut ordered_variants = Vec::new();
+        for lang in PatternLanguage::enumerate() {
+            let lang_name = lang.to_str();
+            if unique_variants.insert(lang_name) {
+                ordered_variants.push(lang_name);
+            }
+            let aliases = lang.get_aliases();
+            for alias in aliases {
+                if unique_variants.insert(alias) {
+                    ordered_variants.push(alias);
+                }
+            }
+            for ext in lang.get_file_extensions() {
+                if unique_variants.insert(ext) {
+                    ordered_variants.push(ext);
+                }
+            }
+        }
+        ordered_variants
+    }
+
+    pub fn value_parser() -> MapValueParser<PossibleValuesParser, fn(String) -> Self> {
+        PossibleValuesParser::new(Self::possible_values()).map(|s| Self::map_value(s))
+    }
+
+    pub fn to_str(&self) -> &'static str {
+        match self {
+            PatternLanguage::JavaScript => "javascript",
+            PatternLanguage::TypeScript => "typescript",
+            PatternLanguage::Tsx => "js",
+            PatternLanguage::Html => "html",
+            PatternLanguage::Css => "css",
+            PatternLanguage::Json => "json",
+            PatternLanguage::Java => "java",
+            PatternLanguage::CSharp => "csharp",
+            PatternLanguage::Python => "python",
+            PatternLanguage::MarkdownBlock => "markdown",
+            PatternLanguage::MarkdownInline => "markdowninline",
+            PatternLanguage::Go => "go",
+            PatternLanguage::Rust => "rust",
+            PatternLanguage::Ruby => "ruby",
+            PatternLanguage::Solidity => "solidity",
+            PatternLanguage::Hcl => "hcl",
+            PatternLanguage::Yaml => "yaml",
+            PatternLanguage::Sql => "sql",
+            PatternLanguage::Vue => "vue",
+            PatternLanguage::Toml => "toml",
+            PatternLanguage::Php => "php",
+            PatternLanguage::PhpOnly => "phponly",
+            PatternLanguage::Universal => "universal",
+        }
     }
 
     pub fn from_string(name: &str, flavor: Option<&str>) -> Option<Self> {
--- a/crates/language/src/target_language.rs
+++ b/crates/language/src/target_language.rs
@@ -47,15 +47,11 @@ use std::path::PathBuf;
 use std::str::FromStr;
 
 #[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
-// #[clap(rename_all = "lower")]
 #[serde(rename_all = "lowercase")]
 pub enum PatternLanguage {
-    // #[value(skip)]
     JavaScript,
-    // #[value(skip)]
     TypeScript,
     #[default]
-    // #[value(name = "js")]
     #[serde(rename = "js")]
     Tsx,
     Html,
@@ -64,9 +60,7 @@ pub enum PatternLanguage {
     Java,
     CSharp,
     Python,
-    // #[value(name = "markdown")]
     MarkdownBlock,
-    // #[value(skip)]
     MarkdownInline,
     Go,
     Rust,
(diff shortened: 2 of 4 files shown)


Alex-ley-scrub:
some screenshots

before:
```
echo 'console.log(hello)' | cargo run -- apply '`hello` => `goodbye`' --stdin --lang=javascript
```
image

after:
```
echo 'console.log(hello)' | cargo run -- apply '`hello` => `goodbye`' --stdin --lang=javascript
```
image
```
echo 'console.log(hello)' | cargo run -- apply '`hello` => `goodbye`' --stdin --lang=javascrip
```
image

crates/language/src/target_language.rs
```diff
@@ -173,6 +185,88 @@ impl PatternLanguage {
...
+            }
+        }
+        ordered_variants
+    }
+
+    pub fn value_parser() -> MapValueParser<PossibleValuesParser, fn(String) -> Self> {
+        PossibleValuesParser::new(Self::possible_values()).map(|s| Self::map_value(s))
+    }
+
+    pub fn to_str(&self) -> &'static str {
```
morgante, line 242: This seems redundant with https://github.com/getgrit/gritql/blob/5e7ca26133fd2d847f73966ca5c45b121b132faf/crates/language/src/target_language.rs#L86. Why do we need to list again?

crates/cli/src/commands/apply_pattern.rs
```diff
@@ -159,7 +159,7 @@ pub struct ApplyPatternArgs {
     #[clap(long)]
     ai: bool,
     /// Change the default language to use for the pattern (if unset, JavaScript is used by default)
-    #[clap(long = "language", alias = "lang")]
+    #[clap(long = "language", alias = "lang", value_parser = PatternLanguage::value_parser())]
```
morgante, line 162: Note that we want to preserve the skipped variants: https://github.com/getgrit/gritql/blob/5e7ca26133fd2d847f73966ca5c45b121b132faf/crates/language/src/target_language.rs#L52

Please add a test to make sure these variants aren't included as options.

crates/language/src/target_language.rs
```diff
@@ -173,6 +185,88 @@ impl PatternLanguage {
...
+    fn get_aliases(&self) -> &'static [&'static str] {
+        for (variant, aliases) in LANGUAGE_ALIASES {
+            if variant == self {
+                return aliases;
+            }
+        }
+        &[]
+    }
+
+    fn map_value(s: String) -> Self {
```
morgante, line 197: I think we could safely just delegate this to from_string and always support aliases inside from_string.

crates/language/src/target_language.rs
```diff
@@ -173,6 +185,88 @@ impl PatternLanguage {
...
+                    return *variant;
+                }
+            }
+        }
+        Self::Universal
+    }
+
+    fn possible_values() -> Vec<&'static str> {
+        let mut unique_variants = std::collections::HashSet::new();
+        let mut ordered_variants = Vec::new();
```
morgante, line 217: Two collections is unnecessary, let's just build the hashset then convert to Vec at the end.

morgante (changes requested):
Thanks, looks like it's on the right track!

crates/language/src/target_language.rs
Alex-ley-scrub: I had used a `HashSet` alone originally like this:

```rust
    fn possible_values() -> Vec<&'static str> {
        let mut unique_variants = std::collections::HashSet::new();
        for lang in PatternLanguage::enumerate() {
            let lang_name = lang.to_str();
            unique_variants.insert(lang_name);
            let aliases = lang.get_aliases();
            for alias in aliases {
                unique_variants.insert(alias);
            }
            for ext in lang.get_file_extensions() {
                unique_variants.insert(ext);
            }
        }
        unique_variants.into_iter().collect()
    }
```

which works but the insertion order is lost and so the order of print out of possible values that is displayed looks weird IMO because it is shuffled. I personally thought that it looked better when the name / aliases / extensions appeared grouped/together like this screenshot I shared [above](https://github.com/getgrit/gritql/pull/544#issuecomment-2425149258):
```
echo 'console.log(hello)' | cargo run -- apply '`hello` => `goodbye`' --stdin --lang=javascrip
```
image

i.e. `js` is grouped with all its aliases/extensions: `javascript, js, jsx, cjs, mjs, vue` and then `ts` / `tsx` is grouped with all its aliases/extensions: `typescript, ts, tsx, cts, mts, flow` and then `rust, rs` appear together, `yaml, yml` appear together etc.

possible values print out (note ordering):
```
error: invalid value 'javascrip' for '--language <LANGUAGE>'
  [possible values: javascript, js, jsx, cjs, mjs, vue, typescript, ts, tsx, cts, mts, flow, html, css, json, java, csharp, cs, python, py, ipynb, markdown, md, mdx, mdoc, markdowninline, go, rust, rs, ruby, rb, solidity, sol, hcl, tf, tfvars, yaml, yml, toml, sql, mysql, postgresql, php, phps, phar, phtml, pht, phponly]

  tip: a similar value exists: 'javascript'
```

which is more similar to how it prints out at the minute before my PR changes:

```
echo 'console.log(hello)' | cargo run -- apply '`hello` => `goodbye`' --stdin --lang=javascript
```
image

possible values print out (note ordering):
```
error: invalid value 'javascript' for '--language <LANGUAGE>'
  [possible values: js, html, css, json, java, csharp, python, markdown, go, rust, ruby, solidity, hcl, yaml, sql, vue, toml, php, phponly]

  tip: a similar value exists: 'java'
```

if we go to the code snippet above, using only a `HashSet`, then it looks like this - which I think personally looks worse but if you prefer that I can certainly make the change:

```
echo 'console.log(hello)' | cargo run -- apply '`hello` => `goodbye`' --stdin --lang=javascrip
```

image

possible values print out (note ordering):
```
error: invalid value 'javascrip' for '--language <LANGUAGE>'
  [possible values: py, ipynb, phar, mts, css, mdx, tsx, mdoc, php, sol, phps, ruby, cts, jsx, md, rs, yaml, solidity, ts, json, yml, tfvars, flow, mysql, tf, vue, js, html, java, hcl, pht, markdowninline, go, postgresql, javascript, csharp, phponly, cs, rb, typescript, python, sql, mjs, markdown, rust, toml, cjs, phtml]

  tip: a similar value exists: 'javascript'
```

i.e. `py` and `ipynb` are together but `python` is "far away". Same for `rust` and `rs` as well as `yaml` and `yml` etc.

but if you are happy with this random ordering / shuffling I will make the change. Just let me know.

crates/language/src/target_language.rs
Alex-ley-scrub: ok can do. I thought it was a bit different because the signature is this and we don't have a flavor when we are parsing CLI args:
```rust
pub fn from_string(name: &str, flavor: Option<&str>) -> Option<Self> {
```

so there may be a slightly different behaviour in terms of parsing preference/order but this probably links to / aligns with your other comments anyway (preserve the skipped variants, get rid of `to_str` and use the `impl fmt::Display for PatternLanguage` etc.)

but I can certainly combine them like I did inside of this `map_value` function and just essentially use None for aliases / extensions:
```rust
        if let Some(lang) = Self::from_string(&s, None) {
            return lang;
        }
```

crates/cli/src/commands/apply_pattern.rs
Alex-ley-scrub: I had added the opposite test by the sounds of it - I had added a test to verify that `javascript` was a valid alias:

https://github.com/getgrit/gritql/blob/a5887a7d44bdc884c94ef076e7edc5d75ad1788d/crates/cli_bin/tests/apply.rs#L2856-L2869

crates/language/src/target_language.rs
Alex-ley-scrub: I guess we don't necessarily - I added it because it was mapping more "literally" - i.e. closely matching the lowercase clap/serde transformation (i.e. `JavaScript => "javascript"`) and any name/renames like (i.e. `Tsx => "js"`, and `MarkdownBlock => "markdown"`) but it wasn't honouring the skipped fields, which you also mentioned. I thought "javascript" was a good alias for "js" so this would be a "default" way to include that alias without explicitly adding it to `LANGUAGE_ALIASES` but perhaps explict is better anyway and I can remove this `to_str` function at the same time?

crates/cli/src/commands/apply_pattern.rs
Alex-ley-scrub: For now I assumed all extensions and all lowercase names of the enum variants should be "by default" become allowed language aliases e.g. `javascript` and `phponly` are valid aliases (of `js` and `php` respectively), but perhaps at least `phponly` is one you wouldn't want to include as it is skipped and it sounds a bit like an internal name rather than a user facing name?

perhaps I remove this `to_str` method and make the aliases more explicit, by adding `javascript` explicitly to `LANGUAGE_ALIASES` (assuming that `javascript` is a desired alias for `js`) but don't add all the lowercase names from the enum variants and honour the skipped variants by using `fmt` instead of `to_str` inside of `possible_values()` to generate the list of possible values, while honouring skipped variants and renamed variants etc.

morgante: Sorry, I should have been clearer:

* `javascript` *should* be a valid alias of `js` so it will work as a CLI option.
* I don't want to *list* all aliases/flavors in the "help" - it makes the help string a lot more complicated / hard to follow. Instead, the help (which I think just comes from possible_values) should only list the primary name for each language.

> so this command should fail? It currently succeeds with my PR changes (from my comment):
No, it should pass.

crates/language/src/target_language.rs
morgante: Yeah, `None` is fine (since aliases should just use the default flavor).

crates/language/src/target_language.rs
morgante: Thanks, it makes sense why you did it this way.

As noted elsewhere, I actually don't want to include all the different flavors in the help. When providing help, providing many different variations of the same underlying value just pollutes the output.

If we switch it to just outputting the primary name for each language the hashset should be fine.

crates/language/src/target_language.rs
morgante: I'm fine with the existing format names (`js`) being used as the names we list in the output so it looks like this can be dropped.

crates/language/src/target_language.rs
Alex-ley-scrub: done in: 0f1b3ce

crates/language/src/target_language.rs
Alex-ley-scrub: removed `to_str()` and replaced its usage with `to_string()` (using `impl fmt::Display for TargetLanguage`)
in: adc787b

crates/cli/src/commands/apply_pattern.rs
Alex-ley-scrub: test added to ensure that `markdowninline` is an invalid alias as it is a skipped variant: 567a8a0

crates/language/src/target_language.rs
Alex-ley-scrub: The way that `PossibleValuesParser` works is different to this I believe ([docs.rs](https://docs.rs/clap/latest/clap/builder/struct.PossibleValuesParser.html)) it (currently) has to display all of them or none of them and any input needs to be one of these possible values but then you can map many-to-one like I had.

I would have to switch to a different approach to achieve what you're asking for and it would be a decent amount of work I think.

probably something like this: https://github.com/clap-rs/clap/issues/4416

i.e. `impl clap::ValueEnum for PatternLanguage` ([docs.rs](https://docs.rs/clap/latest/clap/builder/struct.EnumValueParser.html))

Unless you disagree, I'll try that out and see how hard it is.

morgante: Yeah based on that discussion it looks like `ValueEnum` is what we are looking for. Specifically, allowing the aliases but not displaying them in help.

Alex-ley-scrub: wasn't too bad in the end - a bit more explicit and verbose but very clear what is happening: 20557f2

crates/language/src/target_language.rs
```diff
@@ -118,6 +118,91 @@ impl From<&TargetLanguage> for PatternLanguage {
...
+            Self::Toml,
+            Self::Php,
+            Self::PhpOnly,
+        ]
+    }
+
+    // https://github.com/getgrit/gritql/issues/445
+    // add any more language aliases here
+    // we need to implement the lowercase/rename transformations here ourselves now
+    fn to_possible_value<'a>(&self) -> Option<clap::builder::PossibleValue> {
```
morgante, line 151: This isn't too bad, but I think we could probably avoid manually listing these. Instead it should be possible to use the existing `fmt` implementation to get the default string. Then add aliases by calling `get_file_extensions`.

Alex-ley-scrub: ahh yeah good point that is much cleaner and more concise 💪: e6f5da3

morgante: Looks great, thanks!

morgante (approved):
Looks good, just need to clean up the old code.

crates/language/src/target_language.rs
```diff
@@ -63,9 +64,9 @@ pub enum PatternLanguage {
     Java,
     CSharp,
     Python,
-    #[value(name = "markdown")]
+    // #[value(name = "markdown")]
```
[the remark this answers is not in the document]
Alex-ley-scrub: done! d5a5768

Alex-ley-scrub:
Hi @morgante - is there anything left for me to do before this can be merged? Thanks for your review, the final solution is pretty nice 👌

PR #557, opened 2024-10-23
feat: call built-ins as predicate

arendjr:
I also tried removing `Predicate::Log` in favor of `Predicate::CallBuiltIn`, but this turned out to be more complex, because the log implementation wants to inspect the variable being passed instead of only its value, so I left that as is for now.

Unfortunately, this leaves me without a good test case in this repo. Would you like me to create one?

## Greptile Summary

--


morgante:
> I also tried removing Predicate::Log in favor of Predicate::CallBuiltIn, but this turned out to be more complex, because the log implementation wants to inspect the variable being passed instead of only its value, so I left that as is for now.

We can probably ditch that functionality if it's making the merger harder. I don't consider it critical.

> Unfortunately, this leaves me without a good test case in this repo. Would you like me to create one?

It would probably be good to have a simple one just as a test.

arendjr:
Good news: I was able to move the inspection of the variable to the log compiler instead, so `Predicate::Log` and `Pattern::Log` are now both replaced with a built-in without loss of functionality.

morgante (approved):
Thanks, looks good! Ideally we'd remove the special `log` from the grammar entirely but this is good enough.

Issue #485, opened 2024-09-04
Allow for non-standard suffixes/filenames for languages


dshahbaz:
I have files that should be treated as python, but do not end in `.py`. I can't convince `grit` to respect the files and treat them as python with any option or config value that I've found. Instead, grit fails with:

```
ERROR (code: 410) - Skipped <mypath here> since it is not a python file
```

My workaround is hardlinking <mypath> to <mypath>.py, but it would be ideal to force grit to ignore the suffix, or to register alternative patterns as a file type using `grit.yaml`.

morgante:
I think there's too immediate things we could do:

- Adjust [this logic](https://github.com/getgrit/gritql/blob/44e1536af813f62c5c48dedc781aed56e3089cc9/crates/cli/src/analyze.rs#L208-L218) so that if the path of the file is directly specified (vs. being discovered via traversal) we always attempt to parse it.
- Add an `--extensions` flag to override which extensions are parsed at runtime.

Happy to review a PR.

morgante:
I think option #1 is best.

Alex-ley-scrub:
/attempt #485 

@dshahbaz what is the suffix of the file out of curiosity if you don't mind sharing?

It's not "pyi" by any chance is it? Or another (semi) valid/standard python extension? .pyi - Stub file ([PEP 484](https://peps.python.org/pep-0484/#stub-files))

I only ask because if it is .pyi I believe they are a true subset of .py files (they're just interface files with function signatues and type hints but no implementations, instead just have `...` in the function bodies etc.) and should be able to be parsed just like a normal .py file.

@morgante should we consider to add "pyi" here (I've done that in this commit 2ca176a in case the answer is yes, otherwise can revert):

https://github.com/getgrit/gritql/blob/a1b312bdefed4982f7a96304b88577762f7edcf1/crates/language/src/target_language.rs#L290

https://github.com/getgrit/gritql/blob/a1b312bdefed4982f7a96304b88577762f7edcf1/crates/language/src/target_language.rs#L346

This is my minimal reproducible example (added as a test case in this commit 7fb636d):

These both fail for me using the released version of grit (with or without `--force`):
```
echo 'class MyClass(object): ...' > dummy.nopy
grit apply '`object` => ``' dummy.nopy --lang=python --force
# out: dummy.nopy: ERROR (code: 410) - Skipped dummy.nopy since it is not a python file
# out: Processed 2 files and found 0 matches
```
and
```
[...]

PR #560, opened 2024-10-24
fix: Attempt to analyze file paths in the input without extension check

mobley-trent:
- Attempts to immediately send file paths that are specified in the input to `file_path_tx`

/claim #485

## Greptile Summary

--

### DIFF
--- a/crates/cli/src/analyze.rs
+++ b/crates/cli/src/analyze.rs
@@ -193,9 +193,7 @@ where
     let (found_count, disk_paths) = match my_input {
         ApplyInput::Disk(ref my_input) => {
             let (file_paths_tx, file_paths_rx) = channel();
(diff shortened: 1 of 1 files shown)


mobley-trent:
Hi @morgante is this the correct logic ?

morgante (changes requested):
We don't necessarily want to send them early. We just want to avoid the extension check.

morgante (changes requested):
Still wrong. We don't need to delete the block entirely.

Please do not request again until you have thought through the requirements entirely and added a test case.

mobley-trent:
@morgante maybe I am misunderstanding the problem statement but correct me if I'm wrong. 

We should attempt to send the file if it is found in `my_input.paths`, otherwise we should perform the logging like before, correct ?

PR #562, opened 2024-10-24
fix: 485 allow invalid extensions

### LINKED ISSUE #485: Allow for non-standard suffixes/filenames for languages

Alex-ley-scrub:
/claim #485 

fixes #485 

## Greptile Summary

--


Alex-ley-scrub:
The iteration order over the files isn't stable/consistent so I'll need to change my snapshot not to use 3 files at once:
https://github.com/getgrit/gritql/actions/runs/11508722345/job/32037478385?pr=562#step:9:1068

crates/cli/src/analyze.rs
```diff
@@ -205,14 +205,17 @@ where
                 if file.file_type().unwrap().is_dir() {
                     continue;
                 }
+                if my_input.paths.contains(&file.path().to_path_buf()) {
+                    file_paths_tx.send(file.path().to_path_buf()).unwrap();
+                    continue;
```
morgante, line 210: We can't skip https://github.com/getgrit/gritql/pull/562/files#diff-3bdeb1a4ad73fd2fbf7fc4ced8fd6e1413bfcd3c0fa6508706d21fe45e230f6bR255

morgante (commented):
Snapshots should usually be of the output files + assert the old error doesn't appear.

crates/cli/src/analyze.rs
Alex-ley-scrub: the new line 209 does the same as 255 (the line you linked)? So we're not skipping it exactly, we're just doing it earlier in a sort of guard clause / early exit style. I can't see what purpose 219-254 serves anymore to be honest - but that's just probably because I don't know the codebase well enough yet.

If we now just always try to parse a path if they explicitly passed it in, then why do we need to check if that file doesn't match the chosen language extension(s)?

Alex-ley-scrub: > Snapshots should usually be of the output files + assert the old error doesn't appear.

That makes more sense. Updated those: fabe365

morgante: I'd like to avoid having the same line in 2 places, so try to refactor so we don't have that / line 255 should be the only place we send it.

We do still want to skip cases of non-matching extensions for files that are *not* explicitly passed in, but I think we can drop these lines: https://github.com/getgrit/gritql/pull/562/files#diff-3bdeb1a4ad73fd2fbf7fc4ced8fd6e1413bfcd3c0fa6508706d21fe45e230f6bR220-R252

Alex-ley-scrub: done in these 3 commits: https://github.com/getgrit/gritql/pull/562/files/fabe3652a9e5d12d8ed96c5d15e43a9d5025c83c..52d4952cb96beaa8b1b85ed5a0f94157a64dd0ca

morgante (approved):
Thanks!

Alex-ley-scrub:
@morgante do these bounties get auto awarded once the linked PR merges or not exactly? (This one hasn't yet)

I thought your comment on the issue might automate it but it doesn't seem to have:

https://github.com/getgrit/gritql/issues/485#issuecomment-2436478318

> /assign @Alex-ley-scrub

morgante:
@Alex-ley-scrub No unfortunately I need to manually pay out each bounty.

PR #567, opened 2024-10-27
fix: 539 ensure correctly skipping files based on nested .gitignore files

### LINKED ISSUE #539: .gitignore from root isn't used when running in a subdirectory
It seems we don't properly traverse upwards to pick up multiple .gitignore files when working within subdirectories of the repository.

# Steps to Reproduce:

Consider the following repository structure:

```
/my-repo
├── .gitignore        # Root-level .gitignore
├── src/
│   ├── .gitignore    # Subdirectory-specific .gitignore
│   ├── module1/
│   │   ├── file1.js
│   ├── module2/
│   │   ├── file2.js
└── docs/
    ├── guide.md
```

The .gitignore in the root of the repository might ignore certain file patterns like *.log or .DS_Store:

```
# Contents of /my-repo/.gitignore
*.log
*.tmp
```

The .gitignore inside the src/ subdirectory may have additional exclusions specific to that folder:

```
# Contents of /my-repo/src/.gitignore
*.test.js
/module2/temp/
```

When operating within /src/module1 or /src/module2, the system does not correctly traverse upward to apply both .gitignore files (from /src/.gitignore and /my-repo/.gitignore).

## Expected Behavior:

Grit apply both the .gitignore from the root directory and any additional .gitignore files present in the current directory or its parent directories.
For example, while working in /src/module1, files matching *.log should be ignored from the root .gitignore, and files like *.test.js should be ignored due to the .gitignore in /src/.

This should happen even if my current working directory is `./src/module1`.

## Actual Behavior:

Only the child .gitignore (if any) is considered, and the system fails to pick up rules from higher levels in the directory structure.

This results in files that should be ignored (e.g., .log files) being incorrectly checked by Grit.

## Potential Fix:

Ensure that the .gitignore traversal logic properly checks parent directories and applies all relevant .gitignore rules, as expected in Git's default behavior.

Alex-ley-scrub:
/claim #539 

fixes #539

see https://docs.rs/ignore/latest/ignore/struct.WalkBuilder.html#ignore-rules

## Greptile Summary

--


Alex-ley-scrub:
@morgante - I can't seem to replicate what you are seeing in #539 ? Do you have an example command that fails?

These commands (now in a test) work as I would expect them to? Correct files are ignored.

```
cd src/module1
grit apply '`object` => ``' . --lang=py --force

cd src/module2
grit apply '`object` => ``' . --lang=py --force
```

I think that the `ignore::Walk` already does this (see description on [docs.rs](https://docs.rs/ignore/latest/ignore/struct.WalkBuilder.html#ignore-rules))

I added a test that I think captures the desired behaviour:

https://github.com/getgrit/gritql/blob/a7d5bd32666163fc6069a74cf7ee6efca022f0a4/crates/cli_bin/tests/apply.rs#L2925-L2928

https://github.com/getgrit/gritql/blob/a7d5bd32666163fc6069a74cf7ee6efca022f0a4/crates/cli_bin/tests/apply.rs#L2930-L2938

https://github.com/getgrit/gritql/blob/a7d5bd32666163fc6069a74cf7ee6efca022f0a4/crates/cli_bin/tests/apply.rs#L2959-L2967

image

Alex-ley-scrub:
> I think that the `ignore::Walk` already does this (see description on [docs.rs](https://docs.rs/ignore/latest/ignore/struct.WalkBuilder.html#ignore-rules)) - I do also have some code locally that also manually adds the .gitignore files to the `ignore::Walk` but there's no point pushing it if it is not needed. Although if you can share a command with me that does fail to ignore them - then I'm happy to push it of course.

image

morgante:
Thanks for investigating, it looks like I'm unable to reproduce this too. Sorry for the mixup!

Issue #438, opened 2024-08-01
C# Support


NonlinearFruit:
Supporting C# as a language for matching against would great! There is a lot of language tech debt hanging around from things like:
- .Net Framework to .Net Core to .Net
- Major release versions every year

Microsoft hasn't provided any official migration paths so a tool like gritql would be super useful.

Review [this guide](https://github.com/getgrit/gritql/blob/main/CONTRIBUTING.md#adding-a-new-target-language) and join us on Discord to discuss.

## Acceptance criteria
1. Working grammar + language support
2. At least 10 test cases, including rewrites and metavariables.

morgante:
C# support is definitely interesting for us and planned for this fall.

GTB3NW:
really keen to see c# support. Some competing tools in the same do support it already but gritql looks like the better tool for robust AST s&r - looking forward to seeing this land!

urbit-pilled:
I'm interested in finishing this task (but only if @onyedikachi-david is no longer interested)

morgante:
@onyedikachi-david Can you confirm if you plan to continue?

onyedikachi-david:
Yes I am, i got distracted with other jobs. I'll be wrapping it up soon.

urbit-pilled:
It's been a while since the last signs of progress, so I think it's fair that I jump in and complete this bounty.
/attempt #438

morgante:
Thanks @urbit-pilled this looks good.

morgante:
Algora is broken so I need to aware this manually.

/tip @urbit-pilled $250

PR #585, opened 2024-12-02
feat: Add C# language support

urbit-pilled:
/claim #438 

PS: The `tree-sitter-c-sharp` submodule has introduced a new `tree-sitter-language` crate (https://crates.io/crates/tree-sitter-language) to decouple grammar parsing with the rest of tree-sitter. Unfortunately this is incompatible with `tree-sitter-facade` so I reverted the `tree-sitter-language` changes.

This is why the submodule doesn't point to the official `tree-sitter-c-sharp` repository.


crates/core/src/test.rs
```diff
@@ -15305,3 +15305,789 @@ fn or_file() {
...
+            pattern: r#"
+                |language csharp
+                |
+                |`public class $name {
+                |    public $name($params) {
+                |        $assignments
+                |    }
+                |
+                |    $_
+                |    $_
```
morgante, line 15773: For future tests we usually want to be a bit less strict on shape. I'd personally want to see `$...` work here.

crates/core/src/test.rs
```diff
@@ -15305,3 +15305,789 @@ fn or_file() {
...
+fn csharp_groupby_keyword() {
+    run_test_expected({
+        TestArgExpected {
+            pattern: r#"
+                |language csharp
+                |
+                |`from $item in $collection
+                |group $item by $item.$key` =>
+                |                    `from $item in $collection
+                |                    group $item by $item.$key into g
```
morgante, line 15997: We should also add tests for specific snippets *inside* a linq query.

Ex.

```
language csharp

`group $item by $item.key` => .
```

morgante (approved):
Looks good, would just love a few more tests to award the bounty. Also please run this on a real C# repo and confirm it works.

urbit-pilled:
Thanks for the review @morgante.
I'll add more tests,  test it out on a real repo and add tests within LINQ queries when I get the time (hopefully later today or tomorrow)

crates/core/src/test.rs
urbit-pilled: fixed in cee3e5f98f16f83de3242e944c966dbccfe7f26a

crates/core/src/test.rs
urbit-pilled: fixed in 7ed9e1f30cfd65bc9fb4b5dd6f99b43b4c3300ee

PR #590, opened 2024-12-19
fix JSX string rule

### LINKED ISSUE #552: Strings are not equivalent to JSX strings
This not matching is surprising: https://app.grit.io/studio?key=FYK7zmmQarHSELKAIk21Y

Problem is here: https://github.com/getgrit/gritql/blob/main/resources/metavariable-grammars/javascript-metavariable-grammar.js#L726C1-L743C7

mehul-m-prajapati:
/claim #552

fixes #552


morgante:
Please add a test verifying this actually fixes the issue.

Issue #556, opened 2024-10-23
Elixir support


setop:
I don't know how much effort it cost to support a new language.
But I'm working on a large Elixir code base and I would love to be able to Grit to refactor it reliably.

morgante:
Thanks for your interest! We currently prioritize languages for our commercial customers, but if you'd like to take on adding this yourself the process is described here: https://github.com/getgrit/gritql/blob/main/CONTRIBUTING.md#adding-a-new-target-language

morgante:
The problem is usually [missing contexts](https://github.com/getgrit/gritql/blob/main/CONTRIBUTING.md#snippet-contexts). If you open a draft PR I'm happy to try debugging.

PR #593, opened 2024-12-31
an attempt to add support for elixir

setop:
Towards https://github.com/getgrit/gritql/issues/556


morgante:
Thanks @setop, I took a quick look and pushed some fixes. Here's some nuance:
- It looks like `^` is a reserved symbol in Elixir, so it's probably not a good choice for metavariables. `$` seems like it should work and is our usual default
- In the metavariable grammar, *named* fields will be needed to add for anywhere we want match.

PR #594, opened 2024-12-31
chore: use `&Path` in `FileRegistry::new_from_paths()`

### LINKED ISSUE #591: Improve `FileRegistry` creation APIs
At the moment, [`FileRegistry::new_from_paths`](https://github.com/getgrit/gritql/blob/50b976db637ecbd60daee3f59cc4f868f59bccb0/crates/grit-pattern-matcher/src/pattern/state.rs#L97-L105) has unpleasant signature.

Creating a `Vec<&'a PathBuf>` can be inefficient because it forces consumers to create a vector **of references**, and/or it could be difficult to pass `std` path if the consumers rely on other libraries to handle paths, like [`camino`](https://docs.rs/camino/latest/camino/).

Suggestion: change the API signature to `&[&Path]`, or pass an iterator. 

cc @arendjr

arendjr:
Fixes #591.

Happy New Year!

### DIFF
--- a/crates/grit-pattern-matcher/src/pattern/state.rs
+++ b/crates/grit-pattern-matcher/src/pattern/state.rs
@@ -96,8 +96,7 @@ impl<'a, Q: QueryContext> FileRegistry<'a, Q> {
 
     /// If only the paths are available, create a FileRegistry with empty owners
     /// This is a logic error if you do not later insert the appropriate owners before get_file_owner is called
-    pub fn new_from_paths(file_paths: impl Iterator<Item = &'a Path>) -> Self {
-        let file_paths: Vec<&Path> = file_paths.collect();
+    pub fn new_from_paths(file_paths: Vec<&'a Path>) -> Self {
         Self {
             version_count: file_paths.iter().map(|_| 0).collect(),
             owners: file_paths.iter().map(|_| Vec::new()).collect(),
--- a/crates/grit-pattern-matcher/src/pattern/state.rs
+++ b/crates/grit-pattern-matcher/src/pattern/state.rs
@@ -18,7 +18,7 @@ use grit_util::{
 };
 use rand::SeedableRng;
 use std::ops::Range as StdRange;
-use std::{collections::HashMap, path::PathBuf};
+use std::{collections::HashMap, path::Path};
 
 #[derive(Debug, Clone)]
 pub struct EffectRange<'a, Q: QueryContext> {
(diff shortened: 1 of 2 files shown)


crates/grit-pattern-matcher/src/pattern/state.rs
```diff
@@ -37,7 +37,7 @@ pub struct FileRegistry<'a, Q: QueryContext> {
     /// The number of versions for each file
     version_count: Vec<u16>,
     /// Original file paths, for lazy loading
-    file_paths: Vec<&'a PathBuf>,
+    file_paths: Vec<&'a Path>,
```
arendjr, line 40: Changing it to a `&Path` makes fewer assumptions on where the paths come from, and it doesn't appear to have any downstream effects.

crates/grit-pattern-matcher/src/pattern/state.rs
```diff
@@ -95,8 +95,9 @@ impl<'a, Q: QueryContext> FileRegistry<'a, Q> {
     }
 
     /// If only the paths are available, create a FileRegistry with empty owners
-    /// This is *unsafe* if you do not later insert the appropriate owners before get_file_owner is called
-    pub fn new_from_paths(file_paths: Vec<&'a PathBuf>) -> Self {
+    /// This is a logic error if you do not later insert the appropriate owners before get_file_owner is called
+    pub fn new_from_paths(file_paths: impl Iterator<Item = &'a Path>) -> Self {
+        let file_paths: Vec<&Path> = file_paths.collect();
```
arendjr, line 100: On second thought, I wonder how useful this really is. If we collect into a `Vec` anyway, we may as well let the caller do it, so it's more explicit and we don't hide the allocation.

I'm thinking #591 was actually more about the request to use `&Path` instead of `&PathBuf`, than it was about the change to use an iterator or slice. What do you think, @ematipico?

ematipico: The upstream issue is caused by the fact that we don't use `PathBuf`/`Path` anymore, so when creating a vector of references for the library (this crate), we can't create a Vector of references because they don't live long a enough. 

Passing a vector of references is actually a weird API, because it's a owned collection of references. It doesn't play very well.

So if the caller needs to pass the data, it might as well be a `Vec<PathBuf>` at this point. No lifetimes

arendjr: I meant it more from the consideration that by accepting `&Path` we can already use it more easily with `camino` through the use of `as_std_path()`. The paths can remain references to avoid unnecessary cloning.

You’re right `Vec<PathBuf>` would be easier from a lifetime perspective, but it seems unnecessary from our perspective and it would introduce unnecessary cloning, so I don’t think we need to make that change.

But that leaves the question, do we pass the paths as a `Vec`, a slice, or an iterator. I changed it to an iterator now, but it seems unnecessary: It will be stored in the data structure as a `Vec` anyway, and it’s more idiomatic in Rust to let the caller do the allocation, which gives it the possibility to elide the allocation if an existing variable can be moved instead.

So I don’t actually agree that passing a vector of references is a weird API. It may be less common, but it can still be the right thing to do.

morgante (approved):
Works for me, assuming @ematipico agrees. Happy new year!

ematipico:
That should work, thank you :)

PR #595, opened 2025-01-01
feat: implement `grit format` command

### LINKED ISSUE #574: `grit format` command
Now that biome has added support for Grit as a [formattable language](https://github.com/biomejs/biome/tree/8a90b895663c92aec0697ef4ecb35d15914028cf/crates/biome_grit_formatter) we should add a `grit format` command to the Grit CLI. We can (hopefully) depend on the Biome crate for this.

`grit format` should:
- Look for all grit patterns in the current directory (using our existing mechanisms, like `grit list`), including `grit.yaml`, `*.grit` files, and markdown patterns
- If a `--write` flag is provided, write the re-formatted patterns back. Otherwise just display the diff / changes

Use https://github.com/getgrit/stdlib for a repo to test on, but add at least one CLI test case in https://github.com/getgrit/gritql/tree/main/crates/cli_bin/tests

Arian8j2:
trying to resolve #574
this is just a proof of concept right now and needs more work and time to get complete
/claim #574


morgante:
Thanks @Arian8j2 for working on this, just wanted to chime in that it's definitely on the right track for what I had in mind. Once you're ready, feel free to request a review.

crates/cli/Cargo.toml
```diff
@@ -20,8 +20,9 @@ anyhow = { version = "1.0.70" }
 clap = { version = "4.1.13", features = ["derive"] }
 indicatif = { version = "0.17.5" }
 # Do *NOT* upgrade beyond 1.0.171 until https://github.com/serde-rs/serde/issues/2538 is fixed
```
Arian8j2, line 22: Is this still important because `biome` depends on `serde` version `1.0.217`

crates/cli/src/commands/format.rs
```diff
@@ -0,0 +1,176 @@
...
+    pub write: bool,
+}
+
+pub async fn run_format(arg: &FormatArgs) -> Result<()> {
+    let (mut resolved, _) = resolve_from_cwd(&Source::Local).await?;
+    // sort to have consistent output for tests
+    resolved.sort();
+
+    let file_path_to_resolved = group_resolved_patterns_by_group(resolved);
+    for (file_path, resovled_patterns) in file_path_to_resolved {
```
Arian8j2, line 25: This can be easily executed in parallel, however the tests that verify stdout will fail due to inconsistencies in the output

crates/cli/src/commands/format.rs
```diff
@@ -0,0 +1,176 @@
...
+            "{}:\n{}",
+            file_path.bold(),
+            format_diff(old_file_content, &new_file_content)
+        );
+    }
+
+    Ok(())
+}
+
+fn format_yaml_file(file_content: &str) -> Result<String> {
```
Arian8j2, line 94: Is it ok to also format the whole `yaml` file?

crates/cli/src/commands/format.rs
```diff
@@ -0,0 +1,176 @@
...
+        buffer.push_str(&input[last_ending_byte..hunk.starting_byte]);
+        buffer.push_str(&hunk.new_content);
+        last_ending_byte = hunk.ending_byte;
+
+        if index == hunks.len() - 1 {
+            buffer.push_str(&input[last_ending_byte..]);
+        }
+    }
+    buffer
+}
```
Arian8j2, line 176: this could easily get removed if markdown could only have one grit pattern (and already all examples that i could find in `stdlib` only has one pattern per markdown), but i remember that when parsing markdown it could technically handle multiple patterns, so i just kept it

crates/cli/Cargo.toml
morgante: Let's avoid changing this in the same PR if we can.

crates/cli/src/commands/format.rs
morgante: Instead of changing the input to make tests happy, another option is to sort the test output by line so it is stabilized.

crates/cli/src/commands/format.rs
```diff
@@ -0,0 +1,176 @@
...
+        {
+            resolved_patterns.push(resolved);
+        } else {
+            acc.push((file_path.clone(), vec![resolved]));
+        }
+        acc
+    })
+}
+
+async fn format_file_resovled_patterns(
```
morgante, line 52: typo

crates/cli/src/commands/format.rs
```diff
@@ -0,0 +1,176 @@
...
+    file_path: String,
+    patterns: Vec<ResolvedGritDefinition>,
+    arg: FormatArgs,
+) -> Result<()> {
+    // patterns has atleast one resolve so unwrap is safe
+    let first_pattern = patterns.first().unwrap();
```
morgante, line 58: We're returning a result, so just use `?` instead of unwrap.

crates/cli/src/commands/format.rs
morgante: What's the reason?

crates/cli/src/commands/format.rs
morgante: Some use cases do have multiple patterns (usually for internal use) so we can't assume it.

crates/cli_bin/tests/format.rs
```diff
@@ -0,0 +1,60 @@
...
+    let output = cmd.output()?;
+
+    println!("stderr: {}", String::from_utf8(output.stderr.clone())?);
+    println!("stdout: {}", String::from_utf8(output.stdout.clone())?);
+
+    assert!(
+        output.status.success(),
+        "Command didn't finish successfully"
+    );
+    assert_yaml_snapshot!(String::from_utf8(output.stdout)?);
```
morgante, line 23: Instead of snapshotting the test output, leading to [these problems](https://github.com/getgrit/gritql/pull/595/files#r1902697985), snapshot the final state of the file (after formatting) + assert on test output containing known messages.

crates/cli/src/commands/format.rs
Arian8j2: because we need to parse the whole `yaml` and edit the `body` and then serialize it back to `yaml` format, in this way the `yaml` code also get's formatted. i couldn't find a way to just change `body` field, also because it's usually multi line string it's also hard to just find the bytes and replace it, because we need additional space and tabs to make it valid `yaml` code.

crates/cli_bin/tests/format.rs
Arian8j2: > leading to [these problems](https://github.com/getgrit/gritql/pull/595/files#r1902697985)

what problems? the link just opens pull request file changes.

crates/cli/src/commands/format.rs
Arian8j2: done

crates/cli/src/commands/format.rs
morgante: We should not not serialize/deserialize the whole body. Round-tripping is lossy. We'd also lose any comments in the yaml.

I understand finding and replacing just the bytes is *harder* but it's very much possible.

crates/cli/Cargo.toml
Arian8j2: ok then can i remove that comment to avoid confusion later?

crates/cli_bin/tests/format.rs
Arian8j2: done

crates/cli/src/commands/format.rs
Arian8j2: i could try to find the code like this:
```rust
for i in 1..10 {
    let indent =  " ".repeat(i);
    let grit_code_with_indent = prefix_each_line(indent, grit_code)
    if let Some(pos) = yaml_file_content.position(grit_code_with_indent) {
        // ...
        break
    }
}
```
but it's a bit hacky

morgante: We actually have a lot of utils in GritQL itself for handling indentation—tested [here](https://github.com/getgrit/gritql/blob/0de0cdbe598e987899cfd153b48610682342edc1/crates/cli_bin/tests/apply.rs#L1616)—so I think you should be able to leverage those.

crates/cli/src/commands/format.rs
```diff
@@ -0,0 +1,244 @@
...
+    ending_byte: usize,
+    new_content: String,
+}
+
+/// returns a new string that applies hunk changes
+fn apply_hunk_changes(input: &str, mut hunks: Vec<HunkChange>) -> String {
+    if hunks.is_empty() {
+        return input.to_string();
+    }
+    hunks.sort_by_key(|hunk| hunk.starting_byte);
```
morgante, line 231: Since you're sorting it, just reverse the order (apply last to first) and use replace_range.

crates/cli/src/commands/format.rs
```diff
@@ -0,0 +1,244 @@
...
+        Ok(None)
+    } else {
+        Ok(Some(format_diff(old_file_content, &new_file_content)))
+    }
+}
+
+/// bubble clause that finds a grit pattern with name "\<pattern_name\>" in yaml and
+/// replaces it's body to "\<new_body\>", `format_yaml_file` uses this pattern to replace
+/// pattern bodies with formatted ones
+const YAML_REPLACE_BODY_PATERN: &str = r#"
```
morgante, line 108: Composing a string then re-parsing it is going to be error prone / run into escaping issues.

Better to just build the pattern directly in Rust.

Arian8j2: > Composing a string then parsing it is going to be error prone / run into escaping issues.

Yeah, i was thinking about inserting variable to runtime and use it in the pattern, But couldn't find how to do it

> Better to just build the pattern directly in Rust.
That's gonna be hard :), Is there any place that i can see examples of this?

crates/cli/src/commands/format.rs
Arian8j2: Nice idea, Done

crates/cli/src/commands/format.rs
morgante: > That's gonna be hard :), Is there any place that i can see examples of this?

Unfortunately don't have many examples today, you would have to build it up via the different pattern structs. I'm probably ok merging this as-is.

Arian8j2: > Unfortunately don't have many examples today, you would have to build it up via the different pattern structs.

What about inserting custom meta variable to runtime? Is it easy?

If not i think merging this is ok

morgante: Yeah double quotes should be sufficient.

morgante:
@Arian8j2 Please make sure all tests / clippy pass.

Arian8j2:
@morgante Couldn't build `biome` with current cargo version of CI (1.76.0-nightly):
```
error[E0658]: the `#[expect]` attribute is an experimental feature
   --> .cargo/git/checkouts/biome-d49ce8898b420a50/8bec9fc/crates/biome_string_case/src/lib.rs:629:13
    |
629 |             #[expect(clippy::disallowed_methods)]
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #54503 <https://github.com/rust-lang/rust/issues/54503> for more information
    = help: add `#![feature(lint_reasons)]` to the crate attributes to enable

error[E0658]: the `#[expect]` attribute is an experimental feature
   --> .cargo/git/checkouts/biome-d49ce8898b420a50/8bec9fc/crates/biome_string_case/src/lib.rs:645:13
    |
645 |             #[expect(clippy::disallowed_methods)]
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #54503 <https://github.com/rust-lang/rust/issues/54503> for more information
    = help: add `#![feature(lint_reasons)]` to the crate attributes to enable

error[E0658]: the `#[expect]` attribute is an experimental feature
   --> .cargo/git/checkouts/biome-d49ce8898b420a50/8bec9fc/crates/biome_string_case/src/lib.rs:660:13
    |
660 |             #[expect(clippy::disallowed_methods)]
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #54503 <https://github.com/rust-lang/rust/issues/54503> for more information
    = help: add `#![feature(lint_reasons)]` to the crate attributes to enable
```
Should we perhaps increase the CI rust version?

morgante:
I think newer Rust versions were breaking our binary on osx. Can you see about activating the feature directly?

Arian8j2:
> Can you see about activating the feature directly?

I don't think we can activate feature of other crates without changing them directly, We need to fork and change if we really want to keep the rust version. [stack overflow of same issue](https://stackoverflow.com/questions/46777457/how-do-i-add-feature-to-another-crates-attributes-without-forking-the)

> I think newer Rust versions were breaking our binary on osx

Whats the issue on osx? Maybe fixing that would be easier

crates/cli/Cargo.toml
morgante: Yes please do.

morgante:
@Arian8j2 I'e upgraded the main branch to 1.82, so hopefully you can rebase and this will be mergeable.

morgante:
Looks like you need to run cargo format: https://github.com/getgrit/gritql/actions/runs/12972060346/job/36179168549

Arian8j2:
@morgante Also be careful when including this on release, It still panics on some `stdlib` patterns due to `biome` issues

crates/gritmodule/src/yaml.rs
```diff
@@ -38,10 +39,13 @@ pub fn get_grit_config(source: &str, source_path: &str) -> Result<GritConfig> {
                 pattern_files.push(file);
             }
             crate::config::GritPatternConfig::Pattern(p) => {
-                patterns.push(GritDefinitionConfig::from_serialized(
-                    p,
-                    source_path.to_string(),
-                ));
+                let mut definitation =
```
morgante, line 42:
```suggestion
                let mut definition =
```

morgante:
Hi @Arian8j2 I'm going to take this over. While this initial implementation is kind of okay, it actually has massive memory usage and doesn't work remotely well.

morgante:
Hi @Arian8j2, the main issue I found is explosive memory usage. Running `grit format` on the stdlib uses 25GB of memory.

Obviously we can't ship it in that state. Some of the problems are obvious (we shouldn't keep all patterns in memory at once) but others will require deeper fixes. I started on a rewrite here: https://github.com/getgrit/gritql/pull/606

> Additionally, will the bounty be awarded?

I can only reward it once it's working viably.

Arian8j2:
@morgante
> we shouldn't keep all patterns in memory at once

I don't think patterns take that much memory, Even with `stdlib` that has unusual amount of patterns.
***
I looked into this and the problem is from `biome`, Really small test that reproduces it:
```rust
#[test]
fn biome_memory_issue() {
    let code = "language js\n";
    biome_grit_parser::parse_grit(code);
}
```
Run this test in **release** mode and boom it will explode memory, Interestingly it doesn't explode on debug mode, Previously i just tested the debug version, So i was shocked when you said it has memory issues, One other thing that i found out is that the `\n` has something to do with it, If you remove that it doesn't explode.

morgante:
> I don't think patterns take that much memory, Even with stdlib that has unusual amount of patterns.

We still shouldn't need to keep them all in memory at the same time.

> I looked into this and the problem is from biome, Really small test that reproduces it:
I've opened an issue upstream: https://github.com/biomejs/biome/issues/5032

PR #627, opened 2025-04-04
rm dep: prefix

### LINKED ISSUE #626: Build error in `marzano-core`: dev dependencies cannot be optional
While building for [nixpkgs](https://github.com/NixOS/nixpkgs):

> cargo metadata failure: error: Package `marzano-core v0.2.0 (/build/source/crates/core)` does not have feature `tracing-opentelemetry`. It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.

Rust doesn't allow development dependencies to be optional, see https://github.com/rust-lang/cargo/issues/1596

sarahec:
rust is confusing `opentelemetry-...` as a feature when listed as `dep:opentelemetry-...` in the cli crate. Fix: remove the prefix.

Fixes #626

### DIFF
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -135,12 +135,12 @@ server = [
 ]
 updater = []
 grit_tracing = [
-  "dep:opentelemetry-otlp",
-  "dep:opentelemetry",
-  "dep:opentelemetry_sdk",
-  "dep:tracing-opentelemetry",
-  "dep:tracing-subscriber",
-  "dep:tracing-log",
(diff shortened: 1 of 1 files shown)


morgante:
This doesn't seem quite right. These are dependencies as far as I can tell.

sarahec:
> This doesn't seem quite right. These are dependencies as far as I can tell.

It might not be. The conflict seems to be that tracing-opentelemetry is marked as a dev dependency *and* as an optional dependency (then called in via dep:).

Someone with more rust experience may need to solve #626

Issue #667, opened 2025-11-23
bug: rewrite error when check deletion index


caiuswang:
**Versions**
- gritql: main

**Steps to reproduce**

1. use below file as input
```java
package a.b;

import javax.persistence.Entity;

public class AEntity implements Serializable {
    /**
     * 大大大
     */
    /**
     * 0. 大大 1. 大大大大大 2. 大大大大 3大大大大大大 4. 大大大大.
     */
    /**
     * 0. 大大 1. 大大大大大 2. 大大大大 3. 大大大大大大大 4. 大大大大大大大 5. 大大大大大大大大.
     */
    /**
     * 0. 大大 1. 大大大大 2. 大大大大 3. 大大大大 4. 大大大大 5. 大大大大.
     */
    /**
     * 大大大大大.
     */
    /**
     * 大大大大大大大大.
     */
    @Convert(converter = T.class)
    @Column(name = "messages")
    private Messeges messages;


    @Column(name = "field"                                                             )
    private String field;


    @Column(name = "created_time", nullable = false, length = 19)
    private Date createdTime;
}
```
2. add a java pttern
```grit
pattern test_pattern_rewrite() {
	file($body) where {
		$body <: contains {
			bubble() field_declaration($modifiers, $declarator) where {
				$replaced_ann = "",
				$delete_annotations = [],
				maybe $modifiers <: contains bubble($replaced_ann) annotation(name=$ann_name) as $anno where {
					if ($ann_name <: "Column") {
						if ($replaced_ann <: "") { $replaced_ann = $anno },
						$anno => .
					} else if ($ann_name <: "Convert") {
						if ($replaced_ann <: not "") { $replaced_ann => . },
						$replaced_ann = $anno
					}
				},
				if ($replaced_ann <: not "") {
					$replaced_ann => `@TableField`
				},
				$replaced_ann = ``
			}
		}
	}
}
```
3. apply the pattern, got the error result

<img width="1257" height="481" alt="Image" src="https://github.com/user-attachments/assets/3acbb3d3-a000-4545-ac24-dca123298370" />

```shell
grit apply  test_pattern_rewrite ErrorEntity.java
```

PR #668, opened 2025-11-23
fix(core): correct deletion index calculation for char indices

caiuswang:
## Which problem is this PR solving?
 - this : #667 
- maybe also this :  #416

## Short description of the changes
- error code is below:
image

## Change
- Don't compare char index with byte index—just use char_indices to get the actual byte index instead. 

## How to verify that this has the expected result
add test:  
- rewrite_pattern_correct_with_utf8_content
- rewrite_pattern_remove_dangling_comma_with_utf8_content


morgante (approved):
Looks good, just need to fix the clippy error: https://github.com/honeycombio/gritql/actions/runs/19609104889/job/56742391741?pr=668#step:6:12

PR #690, opened 2026-04-16
chore: some general housekeeping

Netail:
Some general housekeeping

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,0 @@
-* @biomejs/core-contributors @biomejs/maintainers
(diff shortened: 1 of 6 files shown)


.github/CODEOWNERS
ematipico, line 1: I would remove this file if possible, otherwise every member of the team starts to get notifications. Each member can decide how to subscribe to notifications

ematipico (approved):

.github/CODEOWNERS
Netail: We could also make a grit team perhaps 🤔

